PR #45, opened 2022-07-25
update information

invocamanman:
- erase all hardcoded and not used private Keys which are not necessary for testing purposes
- Update deployment scripts


krlosMata (approved):
:smile_cat:

ignasirv (approved):
💯

PR #46, opened 2022-07-26
fix sequencer URL

tclemos:
A new feature implemented in the node requires the Sequencer URL to match the containers we are running


invocamanman (approved):

laisolizq (approved):
:smile_cat:

PR #55, opened 2022-09-05
add real flow test

invocamanman:
- add real flow test
- Fix rebase cahnge which do not add properly the starkInput into the snarkInput hash
- Notice that the real test flow reverts with `('ERC20: transfer to the zero address')` This is due the input has the 0 address for the aggregator address and therefore there's an ERC20 OZ exception about transferring tokens to the address 0, but the goal of the test is to check the correct computation of the input, which passes ( because the tranfer of the tokens happens after it)


krlosMata (approved):
:smile_cat:

ignasirv (approved):

PR #68, opened 2022-11-08
Feature/support dai permit

invocamanman:
- Support DAi permit
- Add permit test for all permit types
- FIx deployment script when mnemonic is not setted


ignasirv (approved):
💯

src/permit-helper.js
```diff
@@ -51,12 +51,125 @@ async function createPermitSignature(tokenContractInstance, wallet, spenderAddre
     return signature;
 }
 
+
+/**
+ * Create a permit signature with the EIP-712 standar
```
krlosMata, line 56:
```suggestion
 * Create a permit signature with DAI standard
```

src/permit-helper.js
```diff
@@ -1,4 +1,4 @@
-const ethers = require('ethers');
+const { ethers } = require('hardhat');
```
krlosMata, line 1: Use `ethers` library is more light rather than `hardhat`
In this way, `hardhat` could be removed from `dependencies` and add it to  `devDependencies` in future iterations

krlosMata: This was changed in https://github.com/0xPolygonHermez/zkevm-contracts/commit/4be88790576949795292010fbf439db2d9f8672b

krlosMata (changes requested):

src/permit-helper.js
```diff
@@ -1,4 +1,4 @@
...
 const { expect } = require('chai');
```
krlosMata, line 2: Add eslint to this file

krlosMata (changes requested):

src/permit-helper.js
invocamanman: i see this was done to use hte provider, but this makes it dififcult to interact otuside of the project, will change it ^^

krlosMata (approved):
:smile_cat:

PR #69, opened 2022-11-16
Add training wheels mechanisms

jesusholoo:
- GlobalExitRootManager
  - Erase `lastGlobalExitRootNum`
  - Now in the `globalExitRootMap` stores the current timestamp instead
  - This timestamp will be used by the bridge by a protection bootstrap mechanism ( `claimTimeout`)
  - If the root does not change, DO NOT update the timestamp and neither emit event of update
  
- Bridge
  - Update the clones method, now with create2 ( for saving 2600 of gas in every call, also seems more clear the SC verification on blockscout), **FE** &#x1F534; should be aware of this, the way that the `precalculatedWrapperAddress` works change a little bit  and depend on the name, symbol and decimals ( they already agree that this is fine tho)
  - Added `claimTimeout`
    - Once claim is call will check that the `globalExtiRoot` of that claim exist and at least `claimTimeout` seconds has passed since that global exit root was submitted
   -  Add `owner` that is able to update the `claimTimeout`
   - Add `emergencyMode` 
     -  Stops all bridges and claims once activated
     -  Only can be activate i deactivated by PoE contract
 
- PoE
  - Add `emergencyMode`
    - Can be activated by a `owner` ( bootstrap mechanism) or by call `proofDifferentState` succesfully
    - Stop `forcedBatches`, `sequenceBatches`, `sequenceForcedBatches` and `verifyBatches` functions
    - Only can be stopped by `securityCouncil` Address
    - Once activate/deactivated also actives/deactives the bridge one emergency one!

- Summary of training wheels mechanism;
  -  Bridge owner can basically stops the claims on the bridge ( setting the `claimTimeout` to max)
  -  PoE owner can basically stops both contracts ( activating the `emergencyMode`)
  - Security council can only interact once emergency mode is active
  - The idea is to remove in some point both ownership's of the contracts
  - Once removed, the only way to go emergency mode is to proof a different state
    - But this means that if there's a bug in the rom that can be exploited to subtract tokens/ether we cannot do anything about it!
    
- Update related tests and deployment scripts

- Add create genesis script

 TODO:
- Create a timelock that will be the proxyManager of all contracts, and reduce the update time if the emergencyMode is active
This was intended to be in this PR but i didint have time sorry xD

### DIFF
--- a/docker/scripts/genesis_docker.json
+++ b/docker/scripts/genesis_docker.json
@@ -1,5 +1,5 @@
 {
-  "root": "0xd95c3987484f77ca1b7d8dc210004b5fb0b5f59186c3b7f02e6e7d5aadd6485f",
+  "root": "0xbf34f9a52a63229e90d1016011655bc12140bba5b771817b88cbf340d08dcbde",
   "genesis": [
     {
       "balance": "100000000000000000000000",
@@ -35,11 +35,6 @@
         "0x0000000000000000000000000000000000000000000000000000000000000002": "0x9d98deabc42dd696deb9e40b4f1cab7ddbf55988"
       },
       "contractName": "GlobalExitRootManagerL2"
-    },
-    {
-      "balance": "0",
-      "nonce": "0",
-      "address": "0x000000000000000000000000000000005ca1ab1e"
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


krlosMata (approved):
:smile_cat:

PR #81, opened 2023-01-19
Feature/add tests

invocamanman:
- Smart contract changes:
  - `deactivateEmergencyState()` had the modifier `ifEmergencyState` duplicated, since the check is done also is the `_deactivateEmergencyState` from `EmergencyManager`
  -  Add the `metadata` to the event `NewWrappedToken` ( Petition of Polygon FE team)
  - Fix view function `verifyMerkleProof` `index` was mistakenly set as `uint64` instead of `uint32`( there was no attack since in the rest of functions  was correctly put as `uint32`)
  - `PolygonZkEVM` don't have to inherit `Initializable` since `OwnableUpgradeable` already inherit it
 
- Add tests
- Add coverage and gas consumption scripts to `package.json`
- Update a `gas_report.md` with gas consumption estimations.
- Coverage is 100% but not all branches are passed though ( some of them are unreachable)
- `Genesis.json` now contains an array of the rawTxs processed


krlosMata (approved):
:smile_cat:

PR #91, opened 2023-02-03
Add fork ID 

invocamanman:
- Rename the function and event `TrustedVerifyBatches` --> `VerifyBatchesTrustedAggregator`
- Add reentrancy check 
- Update solidity version to `0.8.17` that fixes a [solidity compiler bug](https://blog.soliditylang.org/2022/08/08/calldata-tuple-reencoding-head-overflow-bug/)  
  - The affected code would be the `PolygonZkEVM`. This contract itseld is **not** affected by this bug, but if in a future a contract calls the function `sequenceBatches` easily could trigger this error. In order to be more consistent, upgrade the solidity version aswell of this contract
- Added `feeRecipient` on `sequenceBatches` since could be usefull if the trustedSequencer is managed by a Smart Contract
- Add `forkID` parameter, which can be usefull later for future upgrades
  - Add forkID parameter in the input of the zkRom
- Add `UpdateZkEVMVersion` on initialize, this is intended to emit this event in every upgrade of the contract with relevant changes for the sequencers/aggregators


contracts/PolygonZkEVM.sol
```diff
@@ -329,25 +332,38 @@ contract PolygonZkEVM is
         address indexed aggregator
     );
 
+    /**
+     * @dev Emitted everytime the forkID is updated, this includes the first initialization of the contract
+     * This event is inteded to be emitted for every upgrade of the contract with relevant changes for the nodes
```
gpersoon, line 337: typo `inteded` -> `intended`

contracts/lib/DepositContract.sol
```diff
@@ -1,13 +1,13 @@
...
 
-import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
 
 /**
  * This contract will be used as a helper for all the sparse merkle tree related functions
  * Based on the implementation of the deposit eth2.0 contract https://github.com/ethereum/consensus-specs/blob/dev/solidity_deposit_contract/deposit_contract.sol
  */
-contract DepositContract is Initializable {
+contract DepositContract is ReentrancyGuardUpgradeable {
```
gpersoon, line 10: q: why is `ReentrancyGuardUpgradeable` here, while  `__ReentrancyGuard_init()` and usage of `nonReentrant` are in `PolygonZkEVMBridge.sol` ?

invocamanman: It was here just here just to be clear the hierarchy, i want to be the storage slots in that order:
Initializable --> ReentrancyGuardUpgradeable --> DepositContract --> EmergencyManager --> PolygonZkEVMBridge
But i could swap it, it's true, i could do something like:
DepositContract --> Initializable -->  ReentrancyGuardUpgradeable --> EmergencyManager --> PolygonZkEVMBridge

gpersoon: Ok, The only thing I was thinking about: what if the `DepositContract` is reused in the future from another contract and someone forget to put in `__ReentrancyGuard_init()`. Maybe also put a comment about this in the code of `DepositContract`.

deployment/deployContracts.js
```diff
@@ -28,6 +28,8 @@ async function main() {
     const realVerifier = deployParameters.realVerifier || false;
     const { chainID, networkName } = deployParameters;
     const minDelayTimelock = deployParameters.minDelayTimelock || 10; // Should put some default parameter
+    const forkID = 0;
```
krlosMata, line 31: I would set them a config params

krlosMata (changes requested):

krlosMata (approved):
:smile_cat:

contracts/lib/DepositContract.sol
0xleastwood: > it's a goodidea, but, i think with the OZ lib that checks the storage slots when upgrade will be enough

what OZ library checks for storage slots on each upgrade? If it's the proxy contract, I think it only checks a rollback check when the proxy is UUPS. But in your case, transparent upgradeable proxies are used.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Upgrade.sol#L75-L89

0xleastwood: Unless this is what you mean https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/6a1969f54021cbc04287fb81a22449a9348893a7/packages/core/src/storage/index.ts#L22-L37

If I understand this correctly, `upgrades.upgradeProxy()` allows you to assert some new storage layout that is provided by the upgrader as an option?

contracts/PolygonZkEVM.sol
```diff
@@ -425,9 +441,11 @@ contract PolygonZkEVM is
     /**
      * @notice Allows a sequencer to send multiple batches
      * @param batches Struct array which holds the necessary data to append new batches to the sequence
+     * @param feeRecipient Address that will receive the fees from L2
```
0xleastwood, line 444: So I'm guessing when L2 users pay tx fees, the sequencer can opt to be paid out to any arbitrary fee recipient, similar to how L1 works when proposing blocks?

invocamanman: Yes exactly! this becomes even more useful if the trustedSequencer address gets managed by a SC, With some kind of consensus, imagine there are slots of time. Then any sequencer will be able to call sequenceBAtches through the smart contract and choose their fee recipeint

contracts/lib/DepositContract.sol
invocamanman: I din't know exactly, where in the OZ code, but here [upgradeProxy](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#upgrade-proxy) This function does an storage check ( as you can see there's this option unsafeSkipStorageCheck)

ignasirv (approved):
💯

eduadiez (approved):

PR #92, opened 2023-02-08
Feature/new deployment

invocamanman:
- Add deployment script for deploying a deterministic bridge address in both networks
- Add deployer contract
- Add `PolygonZkEVMTestnet` ( the current version for testnet which allow to not allow `forcedBatches`) and set an arbitrary `forceBatchTimeout`


deployment/deploy_parameters.json.example
```diff
@@ -13,6 +13,9 @@
     "admin":"0x617b3a3528F9cDd6630fd3301B9c8911F7Bf063D",
     "trustedAggregator":"0x617b3a3528F9cDd6630fd3301B9c8911F7Bf063D",
     "minDelayTimelock": 3600,
-    "timelockAddress": "0x617b3a3528F9cDd6630fd3301B9c8911F7Bf063D"
-    "forkID": 1
+    "timelockAddress": "0x617b3a3528F9cDd6630fd3301B9c8911F7Bf063D",
+    "pendingStateTimeout": 604799,
+    "trustedAggregatorTimeout": 604799,
+    "forkID": 1,
+    "salt: "0x0000000000000000000000000000000000000000000000000000000000000000"
```
krlosMata, line 20: not valid JSON
```suggestion
    "salt": "0x0000000000000000000000000000000000000000000000000000000000000000"
```

deployment/deployContracts.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable no-await-in-loop */
+/* eslint-disable no-await-in-loop, no-use-before-define, no-lonely-if */
```
krlosMata, line 1: update `README.md`

deployment/deployContracts.js
```diff
@@ -31,8 +31,11 @@ async function main() {
     const forkID = deployParameters.forkID || 0;
     const version = '0.0.1';
 
-    const pendingStateTimeout = deployParameters.pendingStateTimeout || (60 * 60 * 24 * 7 - 1);
-    const trustedAggregatorTimeout = deployParameters.trustedAggregatorTimeout || (60 * 60 * 24 * 7 - 1);
+    const pendingStateTimeout = deployParameters.pendingStateTimeout || (60 * 60 * 24 * 7 - 1); // 1 week minus 1
```
krlosMata, line 34: For local deployment, `npm run deploy:zkEVM:hardhat` a file needed for upgradeability is created `.openzeppelin/unknown-31337.json`.
Second deploy on hardhat does not detect since it looks for `.openzeppelin/harhat.json`.

It does for `goerli`

deployment/deployContracts.js
```diff
@@ -370,3 +411,93 @@ main().catch((e) => {
...
 });
+
+async function deployPolygonZkEVMDeployer(deployer) {
+    const PolgonZKEVMDeployerFactory = await ethers.getContractFactory('PolygonZkEVMDeployer', deployer);
+
+    const deployTxZKEVMDeployer = (PolgonZKEVMDeployerFactory.getDeployTransaction(
+        deployer.address,
+    )).data;
+
+    const gasLimit = ethers.BigNumber.from(1000000); // Put 1 Million, aprox 650k are necessary
```
krlosMata, line 422: I would add this `gasLimit` & `gasPrice` as a constants at the beginning of the script

krlosMata (changes requested):

deployment/deployContracts.js
invocamanman: this is and edge case of hardhat network. Library of OZ do not break with this, bc i think they check if the contract is deployed or not, and redeploy it if necessary. Since is an edge case already managed by the OZ lib, i don't think it's worth to add further checks

deployment/deployContracts.js
invocamanman: I think it's ok this way, i will have to do a different script for mainnet anyway, there i will put in in other format ^^

PR #96, opened 2023-02-21
Feature/optimizations

invocamanman:
- Add hexxens optimization on calculate batch fee
- Update global exit root by sequence batches and not in every deposit if not setted by the user


contracts/PolygonZkEVMBridge.sol
```diff
@@ -129,17 +132,19 @@ contract PolygonZkEVMBridge is
 
     /**
      * @notice Deposit add a new leaf to the merkle tree
-     * @param token Token address, 0 address is reserved for ether
      * @param destinationNetwork Network destination
      * @param destinationAddress Address destination
      * @param amount Amount of tokens
+     * @param token Token address, 0 address is reserved for ether
+     * @param forceUpdateGlobalExitRoot Indicates if the new root is updated or not
```
krlosMata, line 139:
```suggestion
     * @param forceUpdateGlobalExitRoot Indicates if the new global exit root is updated or not
```

contracts/PolygonZkEVM.sol
```diff
@@ -580,6 +580,9 @@ contract PolygonZkEVM is
         // Consolidate pending state if possible
         _tryConsolidatePendingState();
 
+        // Update global exit root if there's new deposits
```
krlosMata, line 583:
```suggestion
        // Update global exit root if there are new deposits
```

krlosMata: This will be always triggered when sequencing.
I would remove comments or just add the logic to update global exit root when there are new deposits

krlosMata: ok...this is handled inside the Bridge xD

krlosMata (changes requested):

krlosMata (approved):
:smile_cat:

PR #101, opened 2023-03-02
Upgradability contract/scripts for forkID

invocamanman:
- Update upgradability scripts
- Update `PolygonZkEVMTestnetV2`
- Check state root inside field on `PolygonZkEVM`


krlosMata (approved):
:smile_cat:

PR #109, opened 2023-03-23
update max transaction length to be compliant with geth

invocamanman:
- update max transaction length to be compliant with geth


krlosMata (approved):
:smile_cat:

PR #122, opened 2023-04-24
Fix main github action

arnaubennassar:
Stop using ok-to-test and remove unnecessary secrets from main action


krlosMata (approved):
:smile_cat:

PR #124, opened 2023-05-23
add package lock json

invocamanman:
- In order to deploy exactly the same version of the contracts, since `@openzeppelin/hardhat-upgrades` install `@openzeppelin/upgrades-core` in a non-stable version it does not deploy exactly the same transparent proxy and therefore the verify script ` verifyMainnetDeployment/verifyMainnetDeployment.js`
Therefore in order to avoid this kind of errors i upload the package-lock.json to have a fixed versions of all the packages


krlosMata (commented):
:smile_cat:

krlosMata (approved):
:smile_cat:

ignasirv (approved):

PR #139, opened 2023-09-17
Fix `undefined` error in 3_deployContracts.js

bap2pecs:
## Problem

I hit the following error running `npx hardhat run deployment/3_deployContracts.js --network goerli`:

```
TypeError: Cannot read properties of undefined (reading 'address')
    at main (/home/ubuntu/zkevm-contracts/deployment/3_deployContracts.js:313:90)
```

## Root Cause Analysis
when L313 is hit, `polygonZkEVMGlobalExitRoot` is undefined. reading the comments, this should read from the `deployment/deploy_ongoing.json` file so here we should use `ongoingDeployment.polygonZkEVMGlobalExitRoot`

### DIFF
--- a/deployment/3_deployContracts.js
+++ b/deployment/3_deployContracts.js
@@ -309,9 +309,8 @@ async function main() {
         ongoingDeployment.polygonZkEVMGlobalExitRoot = polygonZkEVMGlobalExitRoot.address;
         fs.writeFileSync(pathOngoingDeploymentJson, JSON.stringify(ongoingDeployment, null, 1));
     } else {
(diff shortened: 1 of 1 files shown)


invocamanman:
I will close this PR, since the current working branch is on develop and this became deprecated, sorry! ( feel free to reopen it, if you think so)

bap2pecs:
thanks @invocamanman.  I see it's now fixed at https://github.com/0xPolygonHermez/zkevm-contracts/blob/develop/deployment/v2/3_deployContracts.ts#L371

PR #140, opened 2023-09-17
Fix rate limiting in verifyContracts.js

bap2pecs:
I got `Failed to link proxy 0x<REDACTED> with its implementation. Reason: Max rate limit reached` when i run the verify script.

root cause is that Etherscan API has:

image


invocamanman:
Feel free to make similar changes to the current verify scripts, i didn't had the issue tho ^^ ( please make the changes on develop branch)
Thank you so much for your contributions, and sorry

bap2pecs:
@invocamanman believe this is due to I was using a free Etherscan plan which has a quite small rate limit. I will wait until I test out the v2 contracts deployment and see if I still get this issue

PR #147, opened 2023-09-21
Use forkID 5 in deploy_parameters.json.example

bap2pecs:
PUSH0 is used in the contracts but the opcode was only added in the latest prover


bap2pecs:
cc @arnaubennassar

invocamanman:
I will close this PR, since the current working branch is on develop and this became deprecated, sorry! ( feel free to reopen it, if you think so)

PR #174, opened 2024-01-16
Refactor validium naming

arnaubennassar:
### Motivation

Be able to support multiple data availability protocols in a modular way. Each protocol should implement a contract that follows an interface

### Changes

- Renamed contracts/v2/consensus/dataCommittee => contracts/v2/consensus/validium
- PolygonDataCommittee => PolygonValidium
- CDKDataCommittee => PolygonDataCommittee
- Added data availability protocol interface, instead of having the DAC interface, this one it's meant to be generic
- Added constant _PROTOCOL_NAME to PolygonDataCommittee and getter
- Inside PolygonValidium renamed things related to data committee to data availability protocol
- Updated 4_createRollup.ts to be aware that Validium may support different data availabiity protocols


deployment/v2/4_createRollup.ts
```diff
@@ -64,19 +64,28 @@ async function main() {
...
+        "PolygonValidium",
+        "PolygonValidiumEtrog",
     ];
 
     if (!supportedConensus.includes(consensusContract)) {
         throw new Error(`Consensus contract not supported, supported contracts are: ${supportedConensus}`);
     }
 
+    const supporteDataAvailabilityProtocols = [
+        "PolygonDataCommittee",
```
dev1644, line 82: Shouldn't it be `PolygonValidium`? Because `PolygonDataCommittee` is renamed to `PolygonValidium` earlier.

There are other instances where this is being used and also the word `comittee`

invocamanman: hey!
thanks for reviewing the PRs! ^^
It's ok, take in account that this makes refrence to the `supporteDataAvailabilityProtocols` , so which contracts will verify the callData of the Validium
`PolygonValidium` will be the contract where the Batches are sequenced, while `PolygonDataCommittee` the contract that will validate that data  against a data commitee^^

PR #190, opened 2024-02-06
avoid confusion in README.md

bap2pecs:
there is 

> `trustedAggregatorTimeout`: uint64, If a sequence is not verified in this timeout everyone can verify it

so it's confusing if we don't mention that `pendingStateTimeout` means **everyone** can consolidate it after this timeout because for trusted aggregator, it doesn't need to honor this timeout:

```
    function consolidatePendingState(
        uint32 rollupID,
        uint64 pendingStateNum
    ) external {
        RollupData storage rollup = rollupIDToRollupData[rollupID];
        // Check if pending state can be consolidated
        // If trusted aggregator is the sender, do not check the timeout or the emergency state
        if (!hasRole(_TRUSTED_AGGREGATOR_ROLE, msg.sender)) {
            if (isEmergencyState) {
                revert OnlyNotEmergencyState();
            }

            if (!_isPendingStateConsolidable(rollup, pendingStateNum)) {
                revert PendingStateNotConsolidable();
            }
        }
        _consolidatePendingState(rollup, pendingStateNum);
    }
```

### DIFF
--- a/deployment/testV1ToV2/README.md
+++ b/deployment/testV1ToV2/README.md
@@ -30,7 +30,7 @@ deploy:testnet:v1ToV2:localhost
 -   `admin`:address, Admin address, can adjust RollupManager parameters or stop the emergency state
 -   `trustedAggregator`:address, Trusted aggregator address
 -   `trustedAggregatorTimeout`: uint64, If a sequence is not verified in this timeout everyone can verify it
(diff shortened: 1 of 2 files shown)


invocamanman:
Hey!
Pls make all the PRs, on develop, since will be the next "main" branch in short term.

I though it was implied, but ye is more consistent with the comment above ^^

bap2pecs:
@invocamanman just updated!

PR #196, opened 2024-02-19
remove sequence num, for initSequencedBatch

invocamanman:
- remove sequence num
- add initSequencedBatch instead


contracts/v2/interfaces/IPolygonZkEVMVEtrogErrors.sol
```diff
@@ -45,9 +45,9 @@ interface IPolygonZkEVMVEtrogErrors is IPolygonZkEVMErrors {
     error ForceBatchesDecentralized();
 
     /**
-     * @dev Thrown when the sequence number does not match with the one on the contract
+     * @dev Thrown when the last sequenced batch nmber does nto match the init sequeced batch number
```
dev1644, line 48: typo `nmber` & `nto`

PR #208, opened 2024-03-08
feat(v2): add setCustomTokenMapping

pyk:
The existing bridge contract is designed to address two specific scenarios:

1. When the `originToken` originates from the current chain, the contract facilitates the transfer of the escrowed token.
2. When the `originToken` originates from a different chain, the contract first deploys a wrapped version of the token and then mints the wrapped token.

A new requirement has emerged, necessitating the bridge contract to recognize an `existingToken` on the current chain that is valued on a 1:1 basis with the `originToken`.

To accommodate this requirement, this pull request introduces a new function to the bridge contract, named `setTokenWrappedAddress`. This function leverages the existing functionalities of the contract to meet the new requirement efficiently.

The `setTokenWrappedAddress` function enables the rollup manager to:
1. Deploy a contract capable of minting and burning the `existingToken`.
2. Designate the deployed contract as the token's wrapper.

It is important to note that:
- The admin has the authority to update the wrapper contract by invoking `setTokenWrappedAddress` again. This flexibility is crucial for accommodating any changes in the minting/burning logic of the token.

## Flow

Suppose we have

1. Chain A - `firstToken`
2. Chain B - `secondToken` (existing token)

### Step 0: Rollup manager deploy custom wrapper for `secondToken`

Rollup manager deploy custom wrapper that have arbitary logic to mint and burn `secondToken`.

### Step 1: Rollup manager executes `setTokenWrappedAddress`

Rollup manager executes `setTokenWrappedAddress` in Chain B with the following parameters:
- `originNetwork` -> Chain A
- `originTokenAddress` -> `firstToken`
- `wrappedTokenAddress` -> Contract from step 0
- `existingTokenAddress` -> `secondToken`

### Step 2: User bridge from chain A

1. User execute `bridgeAsset` in chain A with `firstToken` as input
2. Bridge contract escrow the `firstToken`
3. `claimAsset` is executed on chain B
4. Using `tokenInfoToWrappedToken` mapping, `secondToken` is minted
5. User receives `secondToken`

### Step 3: User withdraw to chain A

1. User execute `bridgeAsset` in chain B with `secondToken` as input
2. Using `wrappedTokenToTokenInfo` and `existingTokenToWrapper` mapping, custom wrapper contract burn the `secondToken`.
3. `claimAsset` is executed on chain A
4. Bridge contract transfers escrowed token
5. User receives `firstToken`

### DIFF
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -75,6 +75,9 @@ contract PolygonZkEVMBridgeV2 is
     // Wrapped token Address --> Origin token information
     mapping(address => TokenInformation) public wrappedTokenToTokenInfo;
 
+    // Existing token address --> Custom wrapper contract
+    mapping(address => address) public existingTokenToWrapper;
+
     // Rollup manager address, previously PolygonZkEVM
     /// @custom:oz-renamed-from polygonZkEVMaddress
     address public polygonRollupManager;
@@ -248,11 +251,16 @@ contract PolygonZkEVMBridgeV2 is
                 ];
 
                 if (tokenInfo.originTokenAddress != address(0)) {
-                    // The token is a wrapped token from another network
-
-                    // Burn tokens
-                    TokenWrapped(token).burn(msg.sender, amount);
-
+                    // The token is either (1) a wrapped token from another network
+                    // or (2) wrapped with custom contract
+                    address wrapper = existingTokenToWrapper[token];
+                    if (wrapper != address(0)) {
+                        // Burn tokens via custom wrapper via setTokenWrappedAddress
+                        TokenWrapped(wrapper).burn(msg.sender, amount);
+                    } else {
+                        // Burn tokens
+                        TokenWrapped(token).burn(msg.sender, amount);
+                    }
                     originTokenAddress = tokenInfo.originTokenAddress;
                     originNetwork = tokenInfo.originNetwork;
                 } else {
@@ -737,12 +745,20 @@ contract PolygonZkEVMBridgeV2 is
     function setTokenWrappedAddress(
         uint32 originNetwork,
         address originTokenAddress,
-        address wrappedTokenAddress
+        address wrappedTokenAddress,
+        address existingTokenAddress
     ) external onlyRollupManager {
+        // Handle claimAsset on target chain
         bytes32 tokenInfoHash = keccak256(
             abi.encodePacked(originNetwork, originTokenAddress)
         );
         tokenInfoToWrappedToken[tokenInfoHash] = wrappedTokenAddress;
+
+        // Handle bridgeAsset from origin chain
+        wrappedTokenToTokenInfo[
+            existingTokenAddress
+        ] = TokenInformation(originNetwork, originTokenAddress);
+        existingTokenToWrapper[existingTokenAddress] = wrappedTokenAddress;
     }
 
     /**
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -727,6 +727,25 @@ contract PolygonZkEVMBridgeV2 is
     }
 
     /**
+     * @notice Set the address of a wrapper using the token information if already exist
+     * @dev This function is used to allow any existing token to be mapped with
+     *      origin token. Wrapper contract should handle mint/burn of the existing token.
+     * @param originNetwork Origin network
+     * @param originTokenAddress Origin token address, 0 address is reserved for ether
+     * @param wrappedTokenAddress Arbitary contract that implements TokenWrapped interface
+     */
+    function setTokenWrappedAddress(
+        uint32 originNetwork,
+        address originTokenAddress,
+        address wrappedTokenAddress
+    ) external onlyRollupManager {
+        bytes32 tokenInfoHash = keccak256(
+            abi.encodePacked(originNetwork, originTokenAddress)
+        );
+        tokenInfoToWrappedToken[tokenInfoHash] = wrappedTokenAddress;
+    }
+
+    /**
      * @notice Function to activate the emergency state
      " Only can be called by the Polygon ZK-EVM in extreme situations
      */
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -83,6 +83,9 @@ contract PolygonZkEVMBridgeV2 is
     /// @custom:oz-renamed-from polygonZkEVMaddress
     address public polygonRollupManager;
 
+    // Bridge manager address; can set custom mapping for any token
+    address public bridgeManager;
+
     // Native address
     address public gasTokenAddress;
 
@@ -145,6 +148,8 @@ contract PolygonZkEVMBridgeV2 is
      * @param _polygonRollupManager polygonZkEVM address
      * @notice The value of `_polygonRollupManager` on the L2 deployment of the contract will be address(0), so
      * emergency state is not possible for the L2 deployment of the bridge, intentionally
+     * @param _bridgeManager Bridge manager address
+     * @notice Bridge manager can set custom mapping for any token.
      * @param _gasTokenMetadata Abi encoded gas token metadata
      */
     function initialize(
@@ -153,11 +158,13 @@ contract PolygonZkEVMBridgeV2 is
         uint32 _gasTokenNetwork,
         IBasePolygonZkEVMGlobalExitRoot _globalExitRootManager,
         address _polygonRollupManager,
+        address _bridgeManager,
         bytes memory _gasTokenMetadata
     ) external virtual initializer {
         networkID = _networkID;
         globalExitRootManager = _globalExitRootManager;
         polygonRollupManager = _polygonRollupManager;
+        bridgeManager = _bridgeManager;
 
         // Set gas token
         if (_gasTokenAddress == address(0)) {
@@ -187,6 +194,13 @@ contract PolygonZkEVMBridgeV2 is
     modifier onlyRollupManager() {
         if (polygonRollupManager != msg.sender) {
             revert OnlyRollupManager();
+        }
+        _;
+    }
+
+    modifier onlyBridgeManager() {
+        if (bridgeManager != msg.sender) {
+            revert OnlyBridgeManager();
         }
         _;
     }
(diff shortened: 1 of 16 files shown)


contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -726,6 +734,33 @@ contract PolygonZkEVMBridgeV2 is
...
+        // Handle claimAsset on target chain
+        bytes32 tokenInfoHash = keccak256(
+            abi.encodePacked(originNetwork, originTokenAddress)
+        );
+        tokenInfoToWrappedToken[tokenInfoHash] = wrappedTokenAddress;
+
+        // Handle bridgeAsset from origin chain
+        wrappedTokenToTokenInfo[
+            existingTokenAddress
+        ] = TokenInformation(originNetwork, originTokenAddress);
```
omnifient, line 760: if originTokenAddress == address(0), this will make bridgeAsset think that existingTokenAddress is a "native" token from the current chain

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -726,6 +734,33 @@ contract PolygonZkEVMBridgeV2 is
...
+     * @dev This function is used to allow any existing token to be mapped with
+     *      origin token. Wrapper contract should handle mint/burn of the existing token.
+     * @param originNetwork Origin network
+     * @param originTokenAddress Origin token address, 0 address is reserved for ether
+     * @param wrappedTokenAddress Arbitary contract that implements TokenWrapped interface
+     */
+    function setTokenWrappedAddress(
+        uint32 originNetwork,
+        address originTokenAddress,
+        address wrappedTokenAddress,
```
omnifient, line 748: i think this is unnecessary - we just want to set existingTokenAddress
(what would even be here?)

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -726,6 +734,33 @@ contract PolygonZkEVMBridgeV2 is
...
+        address existingTokenAddress
+    ) external onlyRollupManager {
+        // Handle claimAsset on target chain
+        bytes32 tokenInfoHash = keccak256(
+            abi.encodePacked(originNetwork, originTokenAddress)
+        );
+        tokenInfoToWrappedToken[tokenInfoHash] = wrappedTokenAddress;
```
omnifient, line 755: replace wrappedTokenAddress (aka L2 BWUSDC) with existingTokenAddress (aka L2 USDCe)

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -726,6 +734,33 @@ contract PolygonZkEVMBridgeV2 is
...
 
+    /**
+     * @notice Set the address of a wrapper using the token information if already exist
+     * @dev This function is used to allow any existing token to be mapped with
+     *      origin token. Wrapper contract should handle mint/burn of the existing token.
+     * @param originNetwork Origin network
+     * @param originTokenAddress Origin token address, 0 address is reserved for ether
+     * @param wrappedTokenAddress Arbitary contract that implements TokenWrapped interface
+     */
+    function setTokenWrappedAddress(
```
omnifient, line 745: assume

originNetwork == 0
originTokenAddress == L1 USDC
wrappedTokenAddress == ??? maybe L2 bwUSDC
existingTokenAddress == L2 USDCe

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -248,11 +251,16 @@ contract PolygonZkEVMBridgeV2 is
                 ];
 
                 if (tokenInfo.originTokenAddress != address(0)) {
-                    // The token is a wrapped token from another network
-
-                    // Burn tokens
-                    TokenWrapped(token).burn(msg.sender, amount);
-
+                    // The token is either (1) a wrapped token from another network
```
omnifient, line 254: these changes might be unnecessary (or maybe me and paul are missing some obvious which is very possible 😄 )

assume bridgeAsset(L2_USDCe, destination=L1, ...), then token=L2_USDCe

what is wrapper? with the suggested changes below, wrapper==L2_USDCe => wrapper==token

without the changes below, wrapper==wrappedTokenAddress

so what is wrappedTokenAddress? 
if wrappedTokenAddress==L2_BWUSDC, then that's wrong, because it was never minted in the first place 
if wrappedTokenAddress==L2_USDCe, then... wrappedTokenAddress==token

omnifient (commented):
my comments were on the fly while reviewing with Paul, so they might sound a bit confusing.

general take is pattern seems correct (i.e. modifying the tokenInfoToWrappedToken and wrappedTokenToTokenInfo)

contracts/v2/PolygonZkEVMBridgeV2.sol
pyk: > this will make bridgeAsset think that existingTokenAddress is a "native" token from the current chain

Sorry ser, can you give me a link to lines where `bridgeAsset` will think token is native when `originTokenAddres` is zero addy?

AFAIK, `bridgeAsset` will think token address is native when `token` address itself is zero

pyk:
@omnifient sorry ser, I'm having a bit of trouble figuring out exactly which lines need tweaking from the review.

I've put in some initial tests for the scenarios I could come up with. If there's anything I've overlooked, do let me know. Thanks ser! 🙇

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -75,6 +75,9 @@ contract PolygonZkEVMBridgeV2 is
     // Wrapped token Address --> Origin token information
     mapping(address => TokenInformation) public wrappedTokenToTokenInfo;
 
+    // Existing token address --> Custom wrapper contract
+    mapping(address => address) public existingTokenToWrapper;
```
omnifient, line 79:
```suggestion
    mapping(address existingToken => address customWrapper) public existingTokenToWrapper;
```

contracts/v2/PolygonZkEVMBridgeV2.sol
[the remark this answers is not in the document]
pyk: > my main comment is that I don't think you need the separate mapping for the custom tokens (but happy to be proven wrong).

@omnifient your flow above assume that `CT_A` is an existing token in `Ly` and already mintable/burnable by the bridge ser. 

In the discord `@pgpg` says that the custom wrapper can be:
1. the token itself that bridge can mint/burn directly (`A_y`), OR
2. arbitary contract that can mint/escrow/burn existing token (`CT_A` is the wrapper, `A_y` is the token)

CMIIW ([link to the start of the discussion about this](https://discord.com/channels/1037241604068618311/1214799795814662174/1215750743617830922))

So my implementation add the following data to the bridge contract:

```
A -> Origin token in `Lx`
A_y -> Existing token in `Ly`
CT_A -> Arbitary contract in `Ly` that can mint/escrow/burn `A_y`
```

When executing `Ly.bridgeAsset`, user will use `A_y` address as the input hence that's why I need to add the mapping for the custom tokens in order to correctly handle the burning/escrow process via `CT_A`.

--

I will update the function name to `setCustomTokenMapping` ser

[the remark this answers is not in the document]
pyk: > where this potentially needs tweaking is if setCustomTokenMapping() happens AFTER someone had already bridged A and bridge-wrapped A (BW_A) gets created

I don't think we need to handle the migration ser, coz `BW_A` can be bridge back to the `Lx` via `Ly.bridgeAsset` before the `setCustomTokenMapping`, then next bridge procedure will be handled gracefully by the custom mapping.

Do you think we need to `setCustomTokenMapping` revert if `BW_A` (default wrapper) already exists?

omnifient:
lgtm

invocamanman:
Hey, thanks for contributing:

- rollupManager does not exist on L2, therefore this approach cannot be done, also is not clear at all how the information will be forwarded to the L2...
- If we try to do this will be using a different approach.
- Make sure to make the PRs on develop pls

If we add this feature will be after feijoa, so hold on on the PR for a bit, since we are finishing the feijoa realease ( the following 2 weeks)

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -726,6 +735,34 @@ contract PolygonZkEVMBridgeV2 is
...
+     * @param originNetwork Origin network
+     * @param originTokenAddress Origin token address, 0 address is reserved for ether
+     * @param wrappedTokenAddress Arbitary contract that implements TokenWrapped interface
+     */
+    function setCustomTokenMapping(
+        uint32 originNetwork,
+        address originTokenAddress,
+        address wrappedTokenAddress,
+        address existingTokenAddress
+    ) external onlyRollupManager {
```
pgebheim, line 751: We should be using a role just for this. On the L2s there is no rollupManaer deployed.

I think I missed this in the original spec -- but perhaps it's something like a globalRootManager address or something

pgebheim (changes requested):
We need a role just for this operation.

invocamanman:
So, after some conversations i think this PR need the following modifications:

- Add a new role on the bridge smart contract to be able to set the wrapped token `rollupAdmin`.
- This modification should take care of the consensus contracts as well that initialize the bridge injecting a transaction from l1, putting the admin of that network as the `rollupAdmin`


- Fix the current implementation, if im not mistaken a

```
 tokenInfoToWrappedToken[tokenInfoHash] = wrappedTokenAddress;

        // Handle bridgeAsset from origin chain
        wrappedTokenToTokenInfo[existingTokenAddress] = TokenInformation(
            originNetwork,
            originTokenAddress
        );
        existingTokenToWrapper[existingTokenAddress] = wrappedTokenAddress;
```

1. you should set the `wrappedTokenToTokenInfo[wrappedTokenAddress]` instead of the `existingTokenAddress` which should be already set if it was a wrapped of the bridge as i understand the code
2. The logic regarding `existingTokenToWrapper` i think it's not correct. without that mapping it already work, Whats happening now if i'm not mistaken is that when calling bridge using the `wrappedTokenAddress`, the bridge will understand it as if it was a "new token" of this network. Also if called from `existingTokenToWrapper`, `wrappedTokenAddress` will be burned instead, which i think it's very confusing and don't understand the logic behind it

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -726,6 +735,34 @@ contract PolygonZkEVMBridgeV2 is
...
+            abi.encodePacked(originNetwork, originTokenAddress)
+        );
+        tokenInfoToWrappedToken[tokenInfoHash] = wrappedTokenAddress;
+
+        // Handle bridgeAsset from origin chain
+        wrappedTokenToTokenInfo[existingTokenAddress] = TokenInformation(
+            originNetwork,
+            originTokenAddress
+        );
+        existingTokenToWrapper[existingTokenAddress] = wrappedTokenAddress;
```
omnifient, line 763: suggestion: add a comment noting that if `setCustomTokenMapping` is called multiple times for the same `existingTokenAddress`, this will override the previous calls and only keep the last `wrappedTokenAddress`

omnifient:
talked w/ @invocamanman and tldr is:
- we need a new role for `setCustomTokenMapping`
- we're not handling the case where `setCustomTokenMapping` is called >1x for the same `existingTokenAddress` (@pgebheim to decide if we want to handle that or if we just raise awareness in docs)

omnifient:
gm @pyk sorry for the delay:

the role setup is a lot simpler than the rollupManager - you just need to create a role inside the bridge contract
you can revert the changes to the other files, but make sure you update the deployment/initialization scripts

PR #219, opened 2024-03-23
fix: wrong variable name in verify script

JacksonRGB:
output 'polygonRollupManagerAddress'

https://github.com/0xPolygonHermez/zkevm-contracts/blob/v5.0.1-rc.2-fork.8/deployment/v2/3_deployContracts.ts#L532

PR #223, opened 2024-03-26
fix: wrong variable name in verify script 

JacksonRGB:
update from #219 

I am not very skilled about pr. 😑


invocamanman:
nice! thank you :D

PR #246, opened 2024-04-18
Add custom mapping

pyk:
## Summary

- Add new role `bridgeManager` in `PolygonZkEVMBridgeV2`
- `bridgeManager` can map bridged token to existing token in the current chain (e.g. Lx USDCx mapped to existing Ly USDCy)
- User can bridge and wirthdraw mapped token

## Case 1: default wrapper

```
A -> Lx token
DW_A -> Default wrapped token deployed by the bridge in Ly
```

`Ly.bridgeAsset` is executed using `DW_A` directly hence `TokenWrapped(DW_A).burn(msg.sender, amount);`

## Case 2: custom wrapper

```
A -> Lx token
CT_A -> Custom wrapper in Ly
A_y -> Existing token in Ly
```

`Ly.bridgeAsset` is executed using `A_y` as input hence need to call `TokenWrapped(CT_A).burn(msg.sender, amount);`

### DIFF
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -134,6 +134,11 @@ contract PolygonZkEVMBridgeV2 is
     );
 
     /**
+     * @dev Emitted when a bridge manager is updated
+     */
+    event BridgeManagerUpdated(address bridgeManager);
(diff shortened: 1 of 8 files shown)


pyk:
cc @omnifient @pgebheim @invocamanman

PR #249, opened 2024-04-22
Update PolygonZkEVMExistentEtrog.sol

worldmars:
Fix the value of v in the signature


invocamanman:
The transaction is corectly codified, it means to be a pre-eip155 transaction, since it's meant to be used in all chains ( so, in all chainIDs)

PR #257, opened 2024-05-12
[WIP]Add ethda da protocol

x-wagmi:


invocamanman:
so, i would see this contrat as an specific case ( 1 our of 1) of the current SC that manages de data commitee.

PR #285, opened 2024-07-18
Update events for PP

ignasirv:
Add programVKey at `AddNewRollupType` and created new event `VerifyPessimisticTrustedAggregator` instead `VerifyBatchesTrustedAggregator` emitted when the trusted aggregator verifies a pessimistic proof


contracts/v2/PolygonRollupManager.sol
```diff
@@ -292,6 +293,17 @@ contract PolygonRollupManager is
         address indexed aggregator
     );
 
+    /**
+     * @dev Emitted when the trusted aggregator verifies a pessimistic proof
+     */
+    event VerifyPessimisticTrustedAggregator(
```
invocamanman, line 299: same event helps the current bridge service to synchronize everything

ignasirv: Fixed!

krlosMata (approved):
:smile_cat:

docker/scripts/v2/deploy-docker.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-sudo rm -rf docker/gethData/geth_data
```
invocamanman, line 2: sudo might be needed if docker is added at sudo level

krlosMata: added again `sudo`

invocamanman (approved):

invocamanman (approved):

docker/scripts/v2/deploy-docker.sh
ignasirv: I don't think using sudo to run docker compose is a good approach

PR #313, opened 2024-08-21
feat: foundry refactor

JDawg287:
# Description

This PR is a grand effort to shift from the current Hardhat framework as a development toolchain to the Foundry toolchain.

Fixes # SMAR-68

## Changes/Additions

- Since the main focus for the foreseeable future are the `v2` contracts:
  - Moved the contracts inside `./contracts/v2` to the `./contracts` dir
  - Moved rest of the `v1` contracts and the not so used contracts to `./contracts/outdated` dir
- There are a couple of contracts which are difficult to compile while running the `forge coverage` command (even with the `--ir-minimum` flag enabled). They are temporarily removed from the forge build flow by adding a `.ignored` extension at the end of the file names. For example `PolygonZkEVMBridgeV2.sol.ignored`. The originals can be found in the `./contracts-ignored-originals` dir. For the Hardhat compilation an additional compile task is added which overrides the original compile task and renames the "ignored" files to their original name before compiling them and then changing the names back to being ignored. The task can be found in `./tasks/compile.ts` dir.
- Added Foundry tests equivalent of the already existing Hardhat tests for the important contracts in the `test` dir
  - Some tests still utilize the js libraries to generate Merkle proofs so a wrapper is added to expose the lib at `tools/zkevm-commonjs-wrapper.js`
 - The tests also utilize some auto generated deployers generated via the [deployer-kit](https://github.com/0xPolygon/deployer-kit) plugin
 - Added scripts:
   - `script/DeployContracts.s.sol` - to deploy all the entire stack in one go
   - `script/GrantRole.sol` - grant roles for the `PolygonRollupManager` contract via the `PolygonZkEVMTimelock` schedules


eduadiez:
I would like to revive this work. I want us to stop using Hardhat and start using Foundry tooling
cc: @ignasirv @krlosMata @invocamanman

PR #314, opened 2024-08-22
Add rollupManager initializer versioning

ignasirv:


contracts/v2/PolygonRollupManager.sol
```diff
@@ -400,6 +405,13 @@ contract PolygonRollupManager is
         _disableInitializers();
     }
 
+    /**
+     * Initializer function to set new rollup manager version
+     */
+    function initialize() external virtual reinitializer(3) {
+        emit UpdateRollupManagerVersion("banana");
```
invocamanman, line 412: Add constant instead ^^

contracts/v2/PolygonRollupManager.sol
```diff
@@ -231,6 +231,8 @@ contract PolygonRollupManager is
     bytes32 internal constant _EMERGENCY_COUNCIL_ADMIN =
         keccak256("EMERGENCY_COUNCIL_ADMIN");
 
+    string internal constant _ROLLUP_MANAGER_VERSION_PESSIMISTIC = "pessimistic";
```
invocamanman, line 234: maybe it's worth it to have a constnat that always have the same name, and public, to be able to easy retrieval from the node :thinking: something like currentVersion!?¿

contracts/v2/PolygonRollupManager.sol
```diff
@@ -270,6 +270,9 @@ contract PolygonRollupManager is
     // Last timestamp when an aggregation happen
     uint64 public lastAggregationTimestamp;
 
+    // Current rollup manager version
+    string public constant ROLLUP_MANAGER_VERSION = "pessimistic";
```
invocamanman, line 274: move it to constants part of the file^^

invocamanman (approved):

krlosMata (approved):
:smile_cat:

PR #315, opened 2024-08-30
NubitDA integration

justinriema:
### Summary
Development on NubitDA integration as offchain data availability provider.
Corresponding logic has been tested in hardhat docker env.

### Changes
Added NubitDA as offchain data availability provider.
Added Nubit.sol which works as IDataAvailabilityProtocol utilizing SP1 Nuport verification logic (https://github.com/RiemaLabs/sp1-nuport).
Added corresponding configuration for NubitDA logic.


invocamanman:
oki, i didint expec that projects start adding their own DA to our repo, but kinda makes sense.

Let's make this more modular! ^^ instead of copyng half of the DA protocol smart contract, i will create an inheritance contract better. Will do early next week, let me know for when you need this integration ^^

Also some guidelines to make PRs to this repo:

- please do not edit the hardhat-config erasing other networks, it's okay to add your tho
- Do not modify the examples
- Do not change the script to jsut support your network, we could change the "include" condition for example to support both ^^ instead of just supporting yours!!
- Do not commit directly to mainnet but develop!

PR #330, opened 2024-09-16
Sovereign bridge implementation

ignasirv:
Create Bridge and GlobalExitRootManager contracts for Sovereign chains + tests.
`BridgeL2SovereignChain.sol` : extends from `PolygonZkEVMBridgeV2` with the features of:
- Set a bridgeManager address
- Functions to remap and undo remap of token addresses by sovereign token addresses
- Overrided mint/burn functions from BridgeV2 to handle that new remapping

`GlobalExitRootManagerL2SovereignChain.sol` : extends from `PolygonZkEVMGlobalExitRootL2` with the features of:
- New function `insertGlobalExitRoot` only callable by coinbase to update the GER of a Sovereign Chain

### DIFF
--- a/contracts/interfaces/IBasePolygonZkEVMGlobalExitRoot.sol
+++ b/contracts/interfaces/IBasePolygonZkEVMGlobalExitRoot.sol
@@ -9,9 +9,9 @@ interface IBasePolygonZkEVMGlobalExitRoot {
     error OnlyAllowedContracts();
 
     /**
-     * @dev Thrown when the caller is not the coinbase
+     * @dev Thrown when the caller is not the coinbase neither the globalExitRootUpdater
      */
-    error OnlyAggOracleOrCoinbase();
+    error OnlyGlobalExitRootUpdater();
 
     /**
      * @dev Thrown when trying to insert a global exit root that is already set
--- a/contracts/interfaces/IBasePolygonZkEVMGlobalExitRoot.sol
+++ b/contracts/interfaces/IBasePolygonZkEVMGlobalExitRoot.sol
@@ -8,6 +8,26 @@ interface IBasePolygonZkEVMGlobalExitRoot {
      */
     error OnlyAllowedContracts();
 
+    /**
+     * @dev Thrown when the caller is not the coinbase
+     */
+    error OnlyAggOracleOrCoinbase();
+
+    /**
+     * @dev Thrown when trying to insert a global exit root that is already set
+     */
+    error GlobalExitRootAlreadySet();
+
+    /**
+     * @dev Thrown when trying to remove more global exit roots thank inserted
+     */
+    error NotEnoughGlobalExitRootsInserted();
+
+    /**
+     * @dev Thrown when trying to remove a ger that is not the last one
+     */
+    error NotLastInsertedGlobalExitRoot();
+
     function updateExitRoot(bytes32 newRollupExitRoot) external;
 
     function globalExitRootMap(
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -740,7 +740,7 @@ contract PolygonZkEVMBridgeV2 is
      * @notice Function to activate the emergency state
      " Only can be called by the Polygon ZK-EVM in extreme situations
      */
-    function activateEmergencyState() external onlyRollupManager {
+    function activateEmergencyState() external virtual onlyRollupManager {
         _activateEmergencyState();
     }
 
@@ -748,7 +748,7 @@ contract PolygonZkEVMBridgeV2 is
      * @notice Function to deactivate the emergency state
      " Only can be called by the Polygon ZK-EVM
      */
-    function deactivateEmergencyState() external onlyRollupManager {
+    function deactivateEmergencyState() external virtual onlyRollupManager {
         _deactivateEmergencyState();
     }
 
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -89,6 +89,7 @@ contract PolygonZkEVMBridgeV2 is
     bytes public gasTokenMetadata;
 
     // WETH address
+    // @note If WETH address is zero, means the gasToken of the network is ether, else means the network has a custom erc20 gas token. This value is set at initialization of the contract. In case gasTokenAddress != zero, a erc20 contract is deployed and the created address is set as WETH address, to handle wrapped ether
     TokenWrapped public WETHToken;
 
     /**
@@ -442,7 +443,7 @@ contract PolygonZkEVMBridgeV2 is
      * @param mainnetExitRoot Mainnet exit root
      * @param rollupExitRoot Rollup exit root
      * @param originNetwork Origin network
-     * @param originTokenAddress  Origin token address, 0 address is reserved for ether
+     * @param originTokenAddress  Origin token address, 0 address is reserved for gas token address. If WETH address is zero, means this gas token is ether, else means is a custom erc20 gas token
      * @param destinationNetwork Network destination
      * @param destinationAddress Address destination
      * @param amount Amount of tokens
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -92,6 +92,12 @@ contract PolygonZkEVMBridgeV2 is
     TokenWrapped public WETHToken;
 
     /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     */
+    uint256[50] private _gap;
+
+    /**
      * @dev Emitted when bridge assets or messages to another network
      */
     event BridgeEvent(
@@ -238,7 +244,7 @@ contract PolygonZkEVMBridgeV2 is
             // In case ether is the native token, WETHToken will be 0, and the address 0 is already checked
             if (token == address(WETHToken)) {
                 // Burn tokens
-                TokenWrapped(token).burn(msg.sender, amount);
+                _bridgeWrappedAsset(TokenWrapped(token), amount);
 
                 // Both origin network and originTokenAddress will be 0
                 // Metadata will be empty
@@ -250,8 +256,7 @@ contract PolygonZkEVMBridgeV2 is
                 if (tokenInfo.originTokenAddress != address(0)) {
                     // The token is a wrapped token from another network
 
-                    // Burn tokens
-                    TokenWrapped(token).burn(msg.sender, amount);
+                    _bridgeWrappedAsset(TokenWrapped(token), amount);
 
                     originTokenAddress = tokenInfo.originTokenAddress;
                     originNetwork = tokenInfo.originNetwork;
@@ -364,7 +369,7 @@ contract PolygonZkEVMBridgeV2 is
         }
 
         // Burn wETH tokens
-        WETHToken.burn(msg.sender, amountWETH);
+        _bridgeWrappedAsset(WETHToken, amountWETH);
 
         _bridgeMessage(
             destinationNetwork,
@@ -492,7 +497,7 @@ contract PolygonZkEVMBridgeV2 is
                 }
             } else {
                 // Claim wETH
-                WETHToken.mint(destinationAddress, amount);
+                _claimWrappedAsset(WETHToken, destinationAddress, amount);
             }
         } else {
             // Check if it's gas token
@@ -536,7 +541,11 @@ contract PolygonZkEVMBridgeV2 is
                         );
 
                         // Mint tokens for the destination address
-                        newWrappedToken.mint(destinationAddress, amount);
+                        _claimWrappedAsset(
+                            newWrappedToken,
+                            destinationAddress,
+                            amount
+                        );
 
                         // Create mappings
                         tokenInfoToWrappedToken[tokenInfoHash] = address(
@@ -555,7 +564,8 @@ contract PolygonZkEVMBridgeV2 is
                         );
                     } else {
                         // Use the existing wrapped erc20
-                        TokenWrapped(wrappedToken).mint(
+                        _claimWrappedAsset(
+                            TokenWrapped(wrappedToken),
                             destinationAddress,
                             amount
                         );
@@ -608,7 +618,7 @@ contract PolygonZkEVMBridgeV2 is
         address destinationAddress,
         uint256 amount,
         bytes calldata metadata
-    ) external ifNotEmergencyState {
+    ) external virtual ifNotEmergencyState {
         // Destination network must be this networkID
         if (destinationNetwork != networkID) {
             revert DestinationNetworkInvalid();
@@ -646,7 +656,7 @@ contract PolygonZkEVMBridgeV2 is
             );
         } else {
             // Mint wETH tokens
-            WETHToken.mint(destinationAddress, amount);
+            _claimWrappedAsset(WETHToken, destinationAddress, amount);
 
             // Execute message
             /* solhint-disable avoid-low-level-calls */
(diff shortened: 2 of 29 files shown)


contracts/interfaces/IBasePolygonZkEVMGlobalExitRoot.sol
```diff
@@ -8,6 +8,11 @@ interface IBasePolygonZkEVMGlobalExitRoot {
      */
     error OnlyAllowedContracts();
 
+    /**
+     * @dev Thrown when the caller is not the trusted sequencer
```
invocamanman, line 12: comment trusted sequencer

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,205 @@
...
+    function _claimWrappedAsset(
+        TokenWrapped tokenWrapped,
+        address destinationAddress,
+        uint256 amount
+    ) internal override {
+        // If is not mintable transfer instead of mint
+        if (wrappedAddressIsNotMintable[address(tokenWrapped)]) {
+            // Transfer wETH
+            // q: safe transfer?
+            tokenWrapped.transfer(destinationAddress, amount);
```
invocamanman, line 118: safe transfer!

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * @notice Remove the address of a remapped token from the mapping
+     * @notice It also removes the token from the isNotMintable mapping
+     * @param sovereignTokenAddress Address of the sovereign wrapped token
+     */
+    function removeSovereignTokenAddress(
+        address sovereignTokenAddress
+    ) external onlyBridgeManager {
+        delete wrappedTokenToTokenInfo[sovereignTokenAddress];
```
invocamanman, line 188: maybe add check that if this is deleted there's a token with that tokenInfo!? ( so this was remapped already)!

contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol
```diff
@@ -0,0 +1,68 @@
...
+    constructor(
+        address _rollupManager,
+        address _bridgeAddress
+    ) PolygonZkEVMGlobalExitRootV2(_rollupManager, _bridgeAddress) {}
+
+    /**
+     * @notice Update the exit root of one of the networks and the global exit root
+     * @param newRoot new exit tree root
+     */
+    function updateExitRoot(
```
invocamanman, line 49: the logic of udpating the eixt root of this contrac tis already inherit from the globalExitRootVL2, you might have to change the inheritance!

contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: AGPL-3.0
+
+pragma solidity 0.8.20;
+import {PolygonAccessControlUpgradeable} from "../lib/PolygonAccessControlUpgradeable.sol";
```
invocamanman, line 4: remove dependency

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,205 @@
...
+     * @param _bridgeManager bridge manager address
+     */
+    function initialize(
+        uint32 _networkID,
+        address _gasTokenAddress,
+        uint32 _gasTokenNetwork,
+        IBasePolygonZkEVMGlobalExitRoot _globalExitRootManager,
+        address _polygonRollupManager,
+        bytes memory _gasTokenMetadata,
+        address _bridgeManager
```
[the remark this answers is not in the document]
ignasirv: It's initialized at `super.initialize` isn't it?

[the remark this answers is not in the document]
invocamanman: ye, but adding by default a custom implementatino if setted by the admin, i think mihgt be usefull

invocamanman:
jsut run the coverage:
`function setMultipleSovereignTokenAddress` remains untested, even tho will be veery easy to do it so

`migrateLegacyToken` remains untested as wel which i think it's a critical funciton and we should test it!!!

invocamanman:
also we shoudl update the 1_create genesis deployment script to be able to put a flag "sovereign" and deploy these contracts

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -156,7 +156,7 @@ contract PolygonZkEVMBridgeV2 is
         IBasePolygonZkEVMGlobalExitRoot _globalExitRootManager,
         address _polygonRollupManager,
         bytes memory _gasTokenMetadata
-    ) public virtual initializer {
+    ) external virtual initializer {
```
invocamanman, line 159: i know can be external, but thne if we import this bridge into some mock won't compile

invocamanman:
Ger implementation shoould write blockHashes right?¿ probably from AggLayer or L1, to mirror the behaviour of Current Global exit root

Even tho in the zkResidency we though to change that to an index to help us retrieve all the improted GERs, but that depends a lot on what impementaiton we will follow to do that.
For now i wioudl say that you can set block hashes instead of timestamp

invocamanman:
overall LGTM ^^ but we should do a proper review ^^

krlosMata (changes requested):

contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol
```diff
@@ -0,0 +1,41 @@
...
+import "../../PolygonZkEVMGlobalExitRootL2.sol";
+
+/**
+ * Contract responsible for managing the exit roots for the Sovereign chains and global exit roots
+ */
+contract GlobalExitRootManagerL2SovereignChain is PolygonZkEVMGlobalExitRootL2 {
+    /**
+     * @dev Emitted when a new global exit root is inserted
+     */
+    event InsertGlobalExitRoot(bytes32 indexed newGlobalExitRoot);
```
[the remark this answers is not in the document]
krlosMata: I would also add that this functionality will allow to prune the state, so you can better control the increasing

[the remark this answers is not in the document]
krlosMata: @invocamanman thoughts on this ?
I'm thinking this feature as a mechanism in `v0.2.0` aggLayer version to avoid: set wrong GER, users claims from that LER continuously and `pp` cannot be generated.
@ignasirv genesis tool should include a flag (like `sovereignChain: true`) in order to create a genesis with those contracts

deployment/v2/1_createGenesis.ts
```diff
@@ -90,15 +90,18 @@ async function main() {
 
     // Check if it's mainnet deployment
     const isMainnet = deployParameters["isMainnet"] === true ? true : false;
+    let isSovereignChain = deployParameters["isSovereignChain"] === true ? true : false;
 
     if (isMainnet === true) {
+        // If is mainnet, can't be sovereign
```
invocamanman, line 96: the flag is mainnet is regarding to the address used on mainnet, it can be sovereign of course! ^^

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,393 @@
...
+ * Contract responsible to manage the token interactions with other networks
+ */
+contract BridgeL2SovereignChain is
+    PolygonZkEVMBridgeV2,
+    IBridgeL2SovereignChains
+{
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    // Wrapped Token information struct
+    struct SovereignTokenAddress {
```
invocamanman, line 22: i think this struct is overkill since it's just used in the loop call

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,393 @@
...
+     *      origin token.
+     * @notice If this function is called multiple times for the same existingTokenAddress,
+     * this will override the previous calls and only keep the last sovereignTokenAddress.
+     * @notice The tokenInfoToWrappedToken mapping  value is replaced by the new sovereign address but it's not the case for the wrappedTokenToTokenInfo map where the value is added, this way user will always be able to withdraw their tokens
+     * @param originNetwork Origin network
+     * @param originTokenAddress Origin token address, 0 address is reserved for ether
+     * @param sovereignTokenAddress Address of the sovereign wrapped token
+     * @param isNotMintable Flag to indicate if the wrapped token is not mintable
+     */
+    function setSovereignTokenAddress(
```
invocamanman, line 199: i think the sufix address here might be confusing

setSovereignToken?¿

invocamanman: same with other functions/evnets htat has de sufix address

invocamanman:
As an added comment i think the docker does not work since the bytecode of hte new bridge contract is too big.

```
Warning: Contract code size is 31141 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
  --> contracts/v2/sovereignChains/BridgeL2SovereignChain.sol:15:1:
```

We should reduce the bytecode size, the eaisest way is to put it on exceptions on harhdat and lower the optimizer runs by default

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
krlosMata: I guess it adds more bytecode to the SC and more gas consumption. I'm not sure if we can afford this additional cost Vs code-readability.
There are two approaches if we decide to not go with struct that I can think of:
- `bytes` (and do the decoding on the SC)
- `bytes[]` --> so each element is a packed version of the struct
- 4 arrays --> each for each element (sanity check should be done on the length, so all has the same)

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
ignasirv: Mhhh in my opinion it is suitable as we are mainly updating the address of the mapped token (and the mintable flag)

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
invocamanman: i would do 4 arrays ^^

krlosMata (approved):
:smile_cat:

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,397 @@
...
+    /**
+     * @notice Updated bridge manager address
+     * @param _bridgeManager Bridge manager address
+     */
+    function setBridgeManager(
+        address _bridgeManager
+    ) external onlyBridgeManager {
+        if (_bridgeManager == address(0)) revert NotValidBridgeManager();
+        bridgeManager = _bridgeManager;
+        emit BridgeManagerUpdated(bridgeManager);
```
invocamanman, line 157: SetBridgeManager for consistency

Tq101510: > SetBridgeManager for consistency

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
ignasirv: I finally used the struct as it consumes less bytecode of the contract and we are at the limit

krlosMata (approved):
:smile_cat:

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -608,7 +619,7 @@ contract PolygonZkEVMBridgeV2 is
         address destinationAddress,
         uint256 amount,
         bytes calldata metadata
-    ) external ifNotEmergencyState {
+    ) external virtual ifNotEmergencyState {
```
invocamanman, line 622: added virtural for no reason?¿

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,453 @@
...
+            revert OnlyBridgeManager();
+        }
+        _;
+    }
+
+    /**
+     * @notice Updated bridge manager address
+     * @param _bridgeManager Bridge manager address
+     */
+    function setBridgeManager(
```
invocamanman, line 170: settes/getters usually go downside of the code

contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @dev Emitted when the last global exit root is removed
+     */
+    event RemoveLastGlobalExitRoot(bytes32 indexed removedGlobalExitRoot);
+
+    modifier onlyGlobalExitRootUpdater() {
```
invocamanman, line 32: Following the rest of contrats convention, modifiers go after consturctor and initialization

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,453 @@
...
+        } else {
+            // Gas token will be an erc20
+            gasTokenAddress = _gasTokenAddress;
+            gasTokenNetwork = _gasTokenNetwork;
+            gasTokenMetadata = _gasTokenMetadata;
+
+            // Set sovereign weth token or create new if not provided
+            if (_sovereignWETHAddress == address(0)) {
+                // Create a wrapped token for WETH, with salt == 0
+                WETHToken = _deployWrappedToken(
```
invocamanman, line 125: could check tha WETHisNotMIntable is false?¿

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,453 @@
...
+    ) internal override {
+        // If is not mintable transfer instead of mint
+        if (wrappedAddressIsNotMintable[address(tokenWrapped)]) {
+            // Transfer wETH
+            IERC20Upgradeable(address(tokenWrapped)).safeTransfer(
+                destinationAddress,
+                amount
+            );
+        } else {
+            // Claim wETH
```
invocamanman, line 434: comment

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,453 @@
...
+     * @param amount Amount of tokens
+     */
+    function _claimWrappedAsset(
+        TokenWrapped tokenWrapped,
+        address destinationAddress,
+        uint256 amount
+    ) internal override {
+        // If is not mintable transfer instead of mint
+        if (wrappedAddressIsNotMintable[address(tokenWrapped)]) {
+            // Transfer wETH
```
invocamanman, line 428: comment

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -0,0 +1,453 @@
...
+            );
+        } else {
+            // Burn tokens
+            tokenWrapped.burn(msg.sender, amount);
+        }
+    }
+
+    /**
+     * @notice Mints tokens from wrapped token to proceed with the claim, if the token is not mintable it will be transferred
+     * note This function has been extracted to be able to override it by other contracts like Bridge2SovereignChain
```
invocamanman, line 416: commenr

package.json
```diff
@@ -34,7 +34,7 @@
         "@nomicfoundation/hardhat-toolbox": "^3.0.0",
         "@openzeppelin/contracts": "4.8.2",
         "@openzeppelin/contracts-upgradeable": "4.8.2",
-        "@openzeppelin/contracts5": "npm:@openzeppelin/contracts@^5.0.0",
+        "@openzeppelin/contracts5": "npm:@openzeppelin/contracts@5.0.0",
```
invocamanman, line 37: nice! ^^

invocamanman (approved):

contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: AGPL-3.0
+
+pragma solidity 0.8.20;
```
EncryptedPasta, line 3: Solc compiler version 0.8.20 switches the default target EVM version to Shanghai, which means that the generated bytecode will include PUSH0 opcodes. Be sure to select the appropriate EVM version in case you intend to deploy on a chain other than mainnet like L2 chains that may not support PUSH0, otherwise deployment of your contracts will fail.

ignasirv:
Merged at https://github.com/0xPolygonHermez/zkevm-contracts/pull/352

PR #335, opened 2024-09-23
disable initializer=false

krlosMata:
This PR does the following:
- removes `initializer: false` in deployment scripts to call `initialize()` function automatically in `polygonZkEVMGlobalExitRoot` SC


praetoriansentry (approved):
worked well for me!

invocamanman (approved):

PR #339, opened 2024-09-25
Add tests to docker build

ignasirv:
- Added test to docker build network to check build correctness
- New gha only to develop to test docker build gha without pushing the image


docker/scripts/v2/deploy-docker.sh
```diff
@@ -13,4 +15,12 @@ sudo mv deployment/v2/create_rollup_output.json docker/deploymentOutput
 sudo DEV_PERIOD=1 docker-compose -f docker/docker-compose.yml down
 sudo docker build -t hermeznetwork/geth-zkevm-contracts -f docker/Dockerfile .
 # Let it readable for the multiplatform build coming later!
-sudo chmod -R go+rxw docker/gethData
\ No newline at end of file
+sudo chmod -R go+rxw docker/gethData
+# Run container
+sudo docker run -p 8545:8545 -d --name docker_test hermeznetwork/geth-zkevm-contracts
```
invocamanman, line 20: shouldn't this be on another script :thinking: ?¿

.github/workflows/build-docker-test.yml
```diff
@@ -0,0 +1,63 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
krlosMata, line 1: This file has been already created in https://github.com/0xPolygonHermez/zkevm-contracts/blob/feature/ongoingPP/.github/workflows/build-docker.yml where all the unnecessary data from gh secrets has been removed.
I would import that exact file, so it will be easier if we do a rebase afterwards

krlosMata (changes requested):

.github/workflows/build-docker.yml
```diff
@@ -5,56 +5,28 @@ name: Build-Docker image
...
-              uses: docker/setup-buildx-action@v1
-            - name: Build and push
-              uses: docker/build-push-action@v2
-              with:
-                  platforms: linux/amd64,linux/arm64
-                  push: true
-                  tags: hermeznetwork/geth-zkevm-contracts:1.5-integration
-                  file: docker/Dockerfile
-                  context: .
+              run: npm run docker:tests
```
krlosMata, line 32: I see that `npm run docker:tests` has inside the `npm run dockerv2:contracts`.
I would like to run two actions in the gha workflow:
```
name: Build docker
  uses: npm run dockerv2:contracts
name: Test docker
  uses: npm run docker:tests
```

and in the `tests-docker.sh` -->  just run the docker and the tests

In this way, building s testing are totally separated

krlosMata (changes requested):

.github/workflows/build-push-docker.yml
```diff
@@ -0,0 +1,60 @@
...
+            - name: Set up QEMU
+              uses: docker/setup-qemu-action@v1
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v1
+            - name: Build and push
+              uses: docker/build-push-action@v2
+              with:
+                  platforms: linux/amd64,linux/arm64
+                  push: true
+                  tags: hermeznetwork/geth-zkevm-contracts:1.5-integration
```
invocamanman, line 58: jsut double checking the tag, is this correct?¿

ignasirv: I took this file from https://github.com/0xPolygonHermez/zkevm-contracts/blob/feature/ongoingPP/.github/workflows/build-push-docker.yml

krlosMata: No one is using the image created by the gha. Therefore, we can delete this entire gha

krlosMata (approved):
:smile_cat:

invocamanman (approved):

PR #343, opened 2024-10-07
Update e2e with l1InfoTree + upgrade verifier to v2

ignasirv:
- Upgrade SP1 verifier to v2
- Update SP1 e2e test to use l1InfoTree


invocamanman (approved):

contracts/v2/interfaces/ISP1Verifier.sol
```diff
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.20;
 
-// imported from: https://github.com/succinctlabs/sp1-contracts/blob/main/contracts/src/ISP1Verifier.sol
```
krlosMata, line 4: I think we should leave this comment

contracts/verifiers/SP1Verifier.sol
```diff
@@ -4,9 +4,6 @@ pragma solidity ^0.8.20;
 import {ISP1Verifier, ISP1VerifierWithHash} from "../v2/interfaces/ISP1Verifier.sol";
 import {PlonkVerifier} from "./PlonkVerifier.sol";
 
-// Current deployments: https://github.com/succinctlabs/sp1-contracts/tree/main/contracts/deployments
```
krlosMata, line 7: I think we should leave this comment

PR #359, opened 2024-11-18
Sovereign Audit remediations

ignasirv:
The following PR implements de remediations for the audit.
It also fixes a security issue found aposteriori of the audit:
- At sovereign contracts, for the context of a sovereign chain that also has FEP (full execution proofs) there is the possibility to insert arbitrary GER, make claims, and restore previous GER and get the transition proved. 
To avoid this behavior, the following remediations have been implemented:
    - The address that can insert GER and the address that can remove GERs is different, opening the possibility to set zero address to remove GERs functionality, specialy designed for FEP chains
    - A function `unsetMultipleClaimedBitmap` to unset claims from the bitmap has been implemented -> IMPORTANT to audit with careful, it makes tricky bit operations. Can only be called by bridgeManager
    - Many functions from main bridge contract are being overwritten at sovereign contract removing some unnecessary checks int the sovereign context. This has been done to reduce bytecode length to 24kb and make de sc deployable
    - Function `setSovereignTokenAddress` has been removed for bytecode length performance improvement
- Due to this new changes, the bytecode exceeds the limit for a smart contract to be deployed, so some functions from `PolygonZkEVMBRidgeV2` has been overwritten. They have been overwritten without some checks only necessary for bridges that are being deployed on L1, which is not the case.
- The scripts and tests have been updated accordingly


contracts/v2/newDeployments/PolygonRollupManagerNotUpgraded.sol
```diff
@@ -19,8 +19,8 @@ contract PolygonRollupManagerNotUpgraded is PolygonRollupManager {
 
     function initialize(
         address trustedAggregator,
-        uint64 _pendingStateTimeout,
-        uint64 _trustedAggregatorTimeout,
+        uint64 /*_pendingStateTimeout*/,
```
invocamanman, line 22: delete parameters from initialize, instead of comment them ^^

contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -30,16 +30,16 @@ contract PolygonZkEVMBridgeV2 is
...
-    bytes4 private constant _PERMIT_SIGNATURE_DAI = 0x8fcbaf0c;
+    bytes4 internal constant _PERMIT_SIGNATURE_DAI = 0x8fcbaf0c;
 
     // Mainnet identifier
-    uint32 private constant _MAINNET_NETWORK_ID = 0;
+    uint32 internal constant _MAINNET_NETWORK_ID = 0;
 
     // ZkEVM identifier
-    uint32 private constant _ZKEVM_NETWORK_ID = 1;
+    uint32 internal constant _ZKEVM_NETWORK_ID = 1;
```
invocamanman, line 42: i think this both variables are not mandatory to be changed, review and do not update them if not mandatory
also applies `_MAX_LEAFS_PER_NETWORK`

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -378,18 +360,76 @@ contract BridgeL2SovereignChain is
...
+     * @notice Function to check that an index is not claimed and set it as claimed
+     * @dev function override to improve a bit the performance and bytecode not checking unnecessary conditions for sovereign chains context
+     * @param leafIndex Index
+     * @param sourceBridgeNetwork Origin network
+     */
+    function _setAndCheckClaimed(
+        uint32 leafIndex,
+        uint32 sourceBridgeNetwork
+    ) internal override {
+        uint256 globalIndex = uint256(leafIndex) +
```
invocamanman, line 422: join ( in the same SC section) and add comments for all this functions that are overrided to save bytecode

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -440,20 +480,146 @@ contract BridgeL2SovereignChain is
...
+            uint256(sourceBridgeNetwork) *
+            _MAX_LEAFS_PER_NETWORK;
+        (uint256 wordPos, uint256 bitPos) = _bitmapPositions(globalIndex);
+        uint256 mask = ~(1 << bitPos);
+        // Check if the bit is already unset
+        if ((claimedBitMap[wordPos] & (1 << bitPos)) == 0) {
+            revert ClaimNotSet();
+        }
+        // Use bitwise AND with the negated mask to unset the bit
+        claimedBitMap[wordPos] &= mask;
```
invocamanman, line 502: i think the code that we have it's more optimal in terms of gas, and also it's more similar, so easier to audit, i woudl change it for:

  uint256 mask = 1 << bitPos;
        uint256 flipped = claimedBitMap[wordPos] ^= mask;
        if (flipped & mask != 0) {
            revert ClaimNotSet();
        }

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -440,20 +480,146 @@ contract BridgeL2SovereignChain is
...
+    }
+
+    /**
+     * @notice Function to call token permit method of extended ERC20
+     * @dev We override this function from PolygonZkEVMBridgeV2 to improve a bit the performance and bytecode not checking unnecessary conditions for sovereign chains context
+     + @param token ERC20 token address
+     * @param amount Quantity that is expected to be allowed
+     * @param permitData Raw data of the call `permit` of the token
+     */
+    function _permit(
```
invocamanman, line 513: same as above, add comment save bytecode etc..

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
ignasirv: The comment is added at the @dev tag of all those functions that are being overrided

contracts/v2/PolygonZkEVMBridgeV2.sol
ignasirv: You are right that `_MAINNET_NETWORK_ID` and `_ZKEVM_NETWORK_ID` can remain private but others are being used at function overriding from sovereignBridge contract

contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol
```diff
@@ -16,6 +14,10 @@ contract GlobalExitRootManagerL2SovereignChain is
     // globalExitRootUpdater address
     address public globalExitRootUpdater;
 
+    // globalExitRootRemover address
+    // In case of initializing a chain with Full execution proofs, this address should be set to zero, otherwise, some malicious sequencer could insert invalid global exit roots, claim, go back and the execution would be correctly proved.
```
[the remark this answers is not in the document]
ignasirv: What do you mean? I understood that in case of a FEP chain, this address has to be zero. Indeed, if it's not zero can be a multisig but what happens with the comment?

PR #364, opened 2024-12-05
add tool to change minDelay timelock

laisolizq:
This PR includes the tool to change the `minDelay` of the `PolygonZkEVMTimelock` contract.

### DIFF
--- a/tools/changeDelayTimelock/.gitignore
+++ b/tools/changeDelayTimelock/.gitignore
@@ -1,0 +1,2 @@
+change_delay_output.json
+change_delay_timelock.json
--- a/tools/changeDelayTimelock/.gitignore
+++ b/tools/changeDelayTimelock/.gitignore
@@ -1,2 +1,2 @@
-change_delay_output.json
+change_delay_output-*
 change_delay_timelock.json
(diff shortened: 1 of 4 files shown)


tools/changeDelayTimelock/.gitignore
```diff
@@ -0,0 +1,2 @@
+change_delay_output.json
```
krlosMata, line 1: Add timestamp to the output, so it is not overwritten it the tool is launched several times.
example: https://github.com/0xPolygonHermez/zkevm-contracts/blob/v9.0.0-rc.3-pp/tools/addRollupType/addRollupType.ts#L15

tools/changeDelayTimelock/changeDelayTimelock.ts
```diff
@@ -0,0 +1,176 @@
...
+   
+    outputJson.scheduleData = scheduleData;
+    outputJson.executeData = executeData;
+    outputJson.minDelay = timelockDelay;
+    outputJson.functionData = {
+        function: 'updateDelay',
+        parameters: parameters.newMinDelay
+    }
+
+    if(parameters.sendSchedule) {
```
krlosMata, line 121: I would personally not add this functionality to this script.
It will simplify a lot the input parameters and it does make sense for long timelock delays.

Usually, timelock schedule/executions can be done in different ways: metamask (hex data), cast or it could be a `.ts` tool.
This code could be converted into another tool, such like `sendTimelockData` that would be agnostic to the data sent and it could be used for any scheduled/executed timelock data.

Also, removing the ability to send the transaction aligns with the other tools that also creates timelock data

krlosMata (changes requested):

krlosMata (changes requested):

krlosMata (approved):
:smile_cat:

krlosMata (approved):
:smile_cat:

invocamanman (approved):

invocamanman (approved):

PR #382, opened 2025-01-12
refactor(ClaimCompressor): implement dynamic gas limit for bridge calls

VolodymyrBg:
Replace hard-coded 2M gas limit with dynamic gas allocation in ClaimCompressor.sol.
The gas limit is now calculated as half of the remaining gas, capped at 2M.
This change improves gas efficiency while maintaining safety limits.

- Removes TODO comment
- Adds dynamic gas calculation
- Maintains 2M gas upper limit
- Prevents potential out-of-gas scenarios

### DIFF
--- a/contracts/v2/utils/ClaimCompressor.sol
+++ b/contracts/v2/utils/ClaimCompressor.sol
@@ -520,9 +520,14 @@ contract ClaimCompressor {
                     revert(0, 0)
                 }
 
-                // SHould i limit the gas TODO of the call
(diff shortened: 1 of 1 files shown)


invocamanman:
If a call trys to send mroe gas than the context has, does not revert. It automatically imposes a maximum:

```
gas is capped at all but one 64th (remaining_gas / 64) of the remaining gas of the current context. If a call tries to send more, the gas is changed to match the maximum allowed.
```
Found on: https://www.evm.codes/

PR #390, opened 2025-01-27
Feature/create sovereign genesis tool

krlosMata:
This PR does the following:
- standalone tool to generate genesis files for sovereignChains that used vanilla clients


ignasirv (approved):
💯

PR #400, opened 2025-02-11
fix: Improve permit signature validation logic in PolygonZkEVMBridgeV2

crStiv:
Fix logical error in _permit function signature validation:
- Restructure conditional logic for signature checks
- Change nested negative check (sig != _PERMIT_SIGNATURE_DAI) to positive condition
- Move DAI permit validation to dedicated else-if block

Before: Nested negative condition in else block
After: Clear conditional flow with positive signature type checks


invocamanman:
Your are updating a VERY important constant of the contract without a commnet.
I think this is clearly a hack xD

PR #407, opened 2025-02-17
Feature/v0.3.0 testvectors

laisolizq:
- Add utils for encoding functions `aggchainECDSA` & `aggchainFEP`
- Add test-vectors for these utils:
    - `data`: JSONs to generate new test-vectors
    - `generators`: To regenerate test-vectors and check that they match the contracts
    - `finalData`: all infomration after generating test-vectors


ignasirv (approved):
💯

PR #415, opened 2025-02-26
fix: add parentheses in _setAndCheckClaimed condition

VolodymyrBg:
title: fix: add parentheses in _setAndCheckClaimed condition

description: Fixed operator precedence issue in _setAndCheckClaimed function by adding parentheses around bitwise AND operation. The original condition `if (flipped & mask == 0)` was incorrectly evaluating as `if (flipped & (mask == 0))` due to `==` having higher precedence than `&`. Added parentheses to fix the evaluation order: `if ((flipped & mask) == 0)`. This ensures proper checking of claimed indices in the bridge contract.


invocamanman:
Did you check the [cheatsheet](https://docs.soliditylang.org/en/latest/cheatsheet.html)!?
please provdie a test next time to prove your statement

PR #435, opened 2025-04-24
CI: migrate workflows to checkout v4

dizer-ti:
GitHub-hosted runners now use Node 20, so actions/checkout@v4 is required. Workflows only updated—no functional changes.


krlosMata:
base branch should be set to `develop`

PR #438, opened 2025-04-25
fix create sovereign genesis tool

krlosMata:
In `createSovereignGenesis` tool:
- add `emergencyBridgePauser` as an input to the SC
- add `emergencyBridgePauser` in the json output
- 2 spaces on json generated for better readability


ignasirv (approved):
💯

PR #443, opened 2025-05-01
interface token wrapped bridge

krlosMata:
This PR does the following:
- adds minimum erc20 interface to interact with the bridge

### DIFF
--- a/contracts/v2/interfaces/ITokenWrappedBridgeMintable.sol
+++ b/contracts/v2/interfaces/ITokenWrappedBridgeMintable.sol
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.28;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @dev This interface contains all the functions used by the BridgeL2SovereignChain.sol smart contract related to tokens that are mintable
+ * Note that the BridgeL2SovereignChain.sol provides two paths in order to attach a token: token mintable or not mintable
+ * Token mintable: BridgeL2SovereignChain.sol will use the '.mint' and '.burn' functions
+ * Token not mintable: BridgeL2SovereignChain.sol will internally use 'transfer' and 'transferFrom' (This optiosn requires the bridge to have the token assets deposited)
+ */
+interface ITokenWrappedBridgeMintable is IERC20 {
+    /** @dev Creates `amount` tokens and assigns them to `account`, increasing
+     * the total supply.
+     */
+    function mint(address to, uint256 value) external;
+
+    /**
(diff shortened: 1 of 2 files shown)


contracts/v2/interfaces/ITokenWrappedBridgeMintable.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.28;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
ignasirv, line 4: If `@openzeppelin/contracts/` is set, it will pick the OZ version stored in contracts, now is version 5, a few weeks ago was version 4, as it creates a bit of ambiguity I would recommend to explicity specify the version as it is done in the other contracts like: `@openzeppelin/contracts5/`

contracts/v2/interfaces/ITokenWrappedBridgeNotMintable.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.28;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
ignasirv, line 4: Same as other comment

ignasirv (changes requested):

ignasirv:
What is the purpose of this contracts? They are not used in others contracts logic right?

invocamanman:
agree on ignasi, don't get it
- if it's a sovereign remapped/upgradedToken contract can have an arbitrary interface, even tho the most common one will be the OZ one.. 
- This interface does not add Permit interfaces...

I would share our interface `TokenWrapped`  wich i think it's more complete.

krlosMata:
The purpose is to provide the interface of the "aggLayer token" to be shared with other projects so they can integrate with the bridge. Strictly, only `transfer` & `transferFrom` are needed for not mintable tokens (`permit` is optional)  and `burn` & `mint` for mintable).

> If @openzeppelin/contracts/ is set, it will pick the OZ version stored in contracts, now is version 5,

Note that this change does not yet take into account the `upgradeableWrappedTokens` branch.

> What is the purpose of this contracts? They are not used in others contracts logic right?

yep, I do not like it either. 
My initial approach was to provide just on interface to be used, without modify the rest of the SC. We should create the interface for the `wrappedToken` and use that interface on the bridge. Then, share that interface.

krlosMata:
For both...`upgradeableWrappedTokens` & `wrappedToken`

krlosMata:
closing in favor of https://github.com/agglayer/agglayer-contracts/blob/v10.1.0-rc.5/contracts/v2/interfaces/ITokenWrappedBridgeUpgradeable.sol

PR #448, opened 2025-05-14
fix: zkevm-contracts deployment issues

leovct:
## Description

The first [fix](https://github.com/agglayer/agglayer-contracts/commit/18a47dbc752ba4c96ad3434a134e5992260f8368) adds the missing proxiedTokensManager property when creating the sovereign genesis. For reference: https://github.com/0xPolygon/kurtosis-cdk/actions/runs/15025224931/job/42224242830?pr=615#step:7:798

The second [fix](https://github.com/agglayer/agglayer-contracts/commit/9145f53179e7ee410563fa8d6648dc59cace6141) updates the bridge initialization call made in the case of pessimistic chains with non vanilla clients (e.g. cdk-erigon). For reference: https://github.com/0xPolygon/kurtosis-cdk/actions/runs/15022595907/job/42215273757

### DIFF
--- a/deployment/v2/4_createRollup.ts
+++ b/deployment/v2/4_createRollup.ts
@@ -584,7 +584,7 @@ async function main() {
             // Get last GER
             const lastGER = await globalExitRootManagerContract.getLastGlobalExitRoot();
 
-            const dataInjectedTx = await polygonZkEVMBridgeContract.interface.encodeFunctionData("initialize", [
+            const dataInjectedTx = await polygonZkEVMBridgeContract.interface.encodeFunctionData("initialize(uint32,address,uint32,address,address,bytes)", [
                 rollupID,
                 gasTokenAddress,
                 gasTokenNetwork,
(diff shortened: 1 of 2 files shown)


jhkimqd (approved):
+1 lgtm!

leovct:
> Latest development can be found in [this branch](https://github.com/agglayer/agglayer-contracts/tree/feature/wtoken-upgrade) which includes the fix on the [createSovereignGenesis](https://github.com/agglayer/agglayer-contracts/pull/448/files#diff-a249921dded8155d9b4633f3beaca509175626931f9b0b8cd13e405c46697085R191).
> 
> Regarding the `initialize` function, it not necessary to set its params if there is only one `initialize` function. However, I like being more explicit here, so I'd suggest to do this same PR into `feature/wtoken-upgrade` with the `initialize` function change

Ok I'll create a new PR which include the initialize fix then.

It started to fail in kurtosis-cdk because there are two `initialize` methods in the `PolygonZkEVMBridgeV2` contract: 
- https://github.com/agglayer/agglayer-contracts/blob/v10.1.0-rc.1/contracts/v2/PolygonZkEVMBridgeV2.sol#L200
- https://github.com/agglayer/agglayer-contracts/blob/v10.1.0-rc.1/contracts/v2/PolygonZkEVMBridgeV2.sol#L242

leovct:
https://github.com/agglayer/agglayer-contracts/pull/449 cc @krlosMata

PR #450, opened 2025-05-15
separate functionalities

krlosMata:
This PR does the following:
- explicitly separates functionalities to update & unset `defaultAggchainVKeys` in two functions in order to emit en specific event for each one and be less error prone


ignasirv (approved):
💯

laisolizq (approved):
:smile_cat:

invocamanman (approved):

PR #455, opened 2025-05-19
add sortInitParams function

laisolizq:
- Adds a function to sort the `initParams` keys before encoding. Without a consistent order in the JSON, an error would occur during encoding
- Updates the `README` files (which were outdated) to include two missing `initParams` parameters
- Fixes testst: `aggregationVKey` renamed to `aggregationVkey` to match the contract definition and wrong `encodeBytes`


invocamanman (approved):

krlosMata (approved):
:smile_cat:

PR #484, opened 2025-06-19
Feature/aggchain ecsav2

ignasirv:
- Implement new aggchain ECDSAV2 contract
- Added onlyInitializing modifier to internal initialize calls
- Testing for aggchainECDSAV2

### DIFF
--- a/contracts/v2/aggchains/AggchainECDSAV2.sol
+++ b/contracts/v2/aggchains/AggchainECDSAV2.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity 0.8.28;
+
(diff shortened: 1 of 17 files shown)


ignasirv:
> add test to upgrade from PolygonPessimisticConsensus to ecdsav2, i think for we might have problems like initialize, that should be 2, also is a good test to check storage layout

- The test is here: https://github.com/agglayer/agglayer-contracts/pull/484/files#diff-f0e88fc93d444319f782535f0327439ed717943b30c11696237708d54a9cfd14R487

- I dont see why should be initialized after upgrading from PP

ignasirv:
Closed as we will be doing a diferent apprach https://github.com/agglayer/agglayer-contracts/pull/488

PR #485, opened 2025-06-19
add test claim reentrancy

laisolizq:
- Add Bridge Receiver with internal call `.claimMessage`
- Add test with two claimMessage (with reentrancy)


hardhat.config.ts
```diff
@@ -240,6 +240,12 @@ const config: HardhatUserConfig = {
                     metadata: { bytecodeHash: 'none' }, // Get always same bytecode
                 }, // try yul optimizer
             },
+            'contracts/mocks/BridgeMessageReceiverMock.sol': {
+                version: '0.8.28',
+                settings: {
+                    viaIR: true,
```
ignasirv, line 246: What is this config for?

PR #488, opened 2025-06-26
Support to use ALGateway for PP consensus

ignasirv:
Small tweak at rollup manager to use ALGateway to retrieve pp keys for pessimistic chains verifications.

1- Remove condition
````
        if (rollup.rollupVerifierType == VerifierType.ALGateway) {
            // Verify proof. The pessimistic proof selector is attached at the first 4 bytes of the proof
            // proof[0:4]: 4 bytes selector pp
            // proof[4:8]: 4 bytes selector SP1 verifier
            // proof[8:]: proof
            aggLayerGateway.verifyPessimisticProof(
                inputPessimisticBytes,
                proof
            );
        } else {
            // Verify proof
            ISP1Verifier(rollup.verifier).verifyProof(
                rollup.programVKey,
                inputPessimisticBytes,
                proof
            );
        }
````
Set it to 
````
            aggLayerGateway.verifyPessimisticProof(
                inputPessimisticBytes,
                proof
            );
`````
2- Add selector to agglayer node
3- Upgrade rollup manager
Careful: Contract upgrade will have to be coordinated with agglayer upgrade

Advantages:
- Low changes on smart contracts 

Disadvantatges:
- Can't remove pessimistic logic from code
- Contracts upgrade will have to be coordinated with agglayer upgrade
- Agglayer to add a selector on pp proofs
- Need to upgrade rollups

### DIFF
--- a/contracts/v2/PolygonRollupManager.sol
+++ b/contracts/v2/PolygonRollupManager.sol
@@ -239,9 +239,6 @@ contract PolygonRollupManager is
     // Current rollup manager version
     string public constant ROLLUP_MANAGER_VERSION = "al-v0.3.2";
 
-    // default selector for pessimistic proof at ALGateway, used to force pessimistic consensus chains to use pp from ALGateway-
-    bytes4 public constant DEFAULT_PP_SELECTOR = 0xffff0000;
-
     // Hardcoded address used to indicate that this address triggered in an event should not be considered as valid.
     address private constant _NO_ADDRESS =
         0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
@@ -1333,27 +1330,11 @@ contract PolygonRollupManager is
             aggchainData
         );
 
-        if (rollup.rollupVerifierType == VerifierType.ALGateway) {
-            // Verify proof. The pessimistic proof selector is attached at the first 4 bytes of the proof
-            // proof[0:4]: 4 bytes selector pp
-            // proof[4:8]: 4 bytes selector SP1 verifier
-            // proof[8:]: proof
-            aggLayerGateway.verifyPessimisticProof(
-                inputPessimisticBytes,
-                proof
-            );
-        } else {
-            // Append the default PP selector to the proof
-            // proof[0:4]: 4 bytes selector pp
-            bytes memory proofWithPPSelector = abi.encodePacked(
-                DEFAULT_PP_SELECTOR,
-                proof
-            );
-            aggLayerGateway.verifyPessimisticProof(
-                inputPessimisticBytes,
-                proofWithPPSelector
-            );
-        }
+        // Verify proof. The pessimistic proof selector is attached at the first 4 bytes of the proof
+        // proof[0:4]: 4 bytes selector pp
+        // proof[4:8]: 4 bytes selector SP1 verifier
+        // proof[8:]: proof
+        aggLayerGateway.verifyPessimisticProof(inputPessimisticBytes, proof);
 
         // Update aggregation parameters
         lastAggregationTimestamp = uint64(block.timestamp);
--- a/contracts/v2/PolygonRollupManager.sol
+++ b/contracts/v2/PolygonRollupManager.sol
@@ -237,7 +237,10 @@ contract PolygonRollupManager is
         keccak256("EMERGENCY_COUNCIL_ADMIN");
 
     // Current rollup manager version
-    string public constant ROLLUP_MANAGER_VERSION = "al-v0.3.1";
+    string public constant ROLLUP_MANAGER_VERSION = "al-v0.3.2";
+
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: contracts/v2/PolygonRollupManager.sol [hunk 1]
```diff
@@ -237,7 +237,7 @@ contract PolygonRollupManager is
         keccak256("EMERGENCY_COUNCIL_ADMIN");
 
     // Current rollup manager version
-    string public constant ROLLUP_MANAGER_VERSION = "al-v0.3.1";
+    string public constant ROLLUP_MANAGER_VERSION = "al-v0.3.2";
```
krlosMata, line 240: @laisolizq to propose new versioning on the contract.
Contract is already on production, so `v1` is mandatory.
If we consider previous version as `v1.0.1` (the one with `initMigration()`)....this one should be `v1.1.0` since it includes a breaking change for `VerifierType:Pessimistic`

contracts/v2/PolygonRollupManager.sol
```diff
@@ -531,8 +524,12 @@ contract PolygonRollupManager is
         uint32 rollupTypeID = ++rollupTypeCount;
 
         if (rollupVerifierType == VerifierType.Pessimistic) {
-            // No genesis on pessimistic rollups
-            if (genesis != bytes32(0)) revert InvalidRollupType();
+            // Those params should be zero for Pessimistic rollup types
+            if (
+                genesis != bytes32(0) ||
+                verifier != address(0) ||
+                programVKey != bytes32(0)
```
krlosMata, line 531: we can remove the `programVKey` from the inputs

contracts/v2/PolygonRollupManager.sol
```diff
@@ -531,8 +524,12 @@ contract PolygonRollupManager is
         uint32 rollupTypeID = ++rollupTypeCount;
 
         if (rollupVerifierType == VerifierType.Pessimistic) {
-            // No genesis on pessimistic rollups
-            if (genesis != bytes32(0)) revert InvalidRollupType();
+            // Those params should be zero for Pessimistic rollup types
+            if (
+                genesis != bytes32(0) ||
```
krlosMata, line 529: `forkID` also to be 0.
Check are the same as `ALGateway`. Therefore, it could be done together.

contracts/v2/PolygonRollupManager.sol
```diff
@@ -492,9 +487,7 @@ contract PolygonRollupManager is
     /**
      * Initializer function to set new rollup manager version
      */
-    function initialize() external virtual reinitializer(5) {
-        emit UpdateRollupManagerVersion(ROLLUP_MANAGER_VERSION);
-    }
+    function initialize() external virtual reinitializer(5) {}
```
krlosMata, line 490: we can remove the `initialize` then

krlosMata (changes requested):

contracts/v2/PolygonRollupManager.sol
ignasirv: Decided to remove in next upgrade among other unused params

contracts/v2/PolygonRollupManager.sol
laisolizq: My proposal is that the current version should be `v2.1.0`, so with this change it would be `v2.1.1`.
You can see my proposal here: https://github.com/agglayer/agglayer-contracts/pull/502

ignasirv: This PR will be merged before the versioning restrictions so they will be applied on next updates

contracts/v2/PolygonRollupManager.sol
```diff
@@ -492,9 +487,8 @@ contract PolygonRollupManager is
     /**
      * Initializer function to set new rollup manager version
      */
-    function initialize() external virtual reinitializer(5) {
-        emit UpdateRollupManagerVersion(ROLLUP_MANAGER_VERSION);
-    }
+    /// @dev Currently unused , keeping it here for keeping track of the last used reinitializer version
+    // function initialize() external virtual reinitializer(5) {}
```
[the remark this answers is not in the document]
ignasirv: In other versions, the initialize function was just triggering an event with the new version.
I think it's not necessary as when upgrading there is already an upgrade event and the version is already handled by the contract itself. 
Had no sense to do this "versioning" management in this contract and not in others, was increasing the initializer version without sense.

krlosMata (approved):
😸

laisolizq (approved):
:sparkles:

contracts/v2/PolygonRollupManager.sol
ignasirv: It's already added

ignasirv:
Closed in favor to upgrade directly to new multisig aggchain type

PR #499, opened 2025-07-11
agg oracle comittee

invocamanman:
- Add agg oracle comittee
- Middle ware contract, that aims to be hte `globalExitRootUpdater` on `GlobalExitRootManagerL2SovereignChain`
- It has multiple oracle members, that once reaches quoruom, the GER is consoldiated, and inserted on the  `GlobalExitRootManagerL2SovereignChain`

specs: https://github.com/agglayer/specs/pull/38


contracts/v2/sovereignChains/AggOracleManager.sol
```diff
@@ -0,0 +1,307 @@
+// SPDX-License-Identifier: AGPL-3.0
+
+pragma solidity 0.8.28;
+
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable5/access/OwnableUpgradeable.sol";
+import {GlobalExitRootManagerL2SovereignChain} from "./GlobalExitRootManagerL2SovereignChain.sol";
```
ignasirv, line 6: Not an issue at all but is already imported in the Interface.
- Consider adding `insertGlobalExitRoot` in the `IGlobalExitRootManagerL2SovereignChain` and import the interface
- Or, return address in the immutable instead of`GlobalExitRootManagerL2SovereignChain`

contracts/v2/sovereignChains/AggOracleManager.sol
```diff
@@ -0,0 +1,307 @@
...
+     * @notice Remove an oracle member.
+     * Only the owner can call this function.
+     * @param oracleMemberAddress Address of the oracle member to remove
+     * @param oracleMemberIndex Index of the oracle member to remove
+     */
+    function removeOracleMember(
+        address oracleMemberAddress,
+        uint256 oracleMemberIndex
+    ) external onlyOwner {
+        bytes32 lastVotedReportHash = addressToLastProposedGER[
```
ignasirv, line 204: The name `lastVotedReportHash` is a bit confusing, maybe rename it to `lastProposedGER`

alrevuelta (commented):
(for got to submit the comments from lasts week, a bit late)

contracts/v2/sovereignChains/AggOracleManager.sol
ignasirv: This way you will not need to import a contract (GlobalExitRootManagerL2SovereignChain) in an Interface (IAggOracleManager).
I mean from:
`GlobalExitRootManagerL2SovereignChain public immutable globalExitRootManagerL2Sovereign;`
to
`address public immutable globalExitRootManagerL2Sovereign;`

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,311 @@
...
+    // Oracle functions
+    ///////////////////
+
+    /**
+     * @notice Propose a global exit root.
+     * This function can only be called by an oracle member.
+     * If the quorum is reached, the GER is consolidated.
+     * @param proposedGlobalExitRoot Global exit root proposed
+     */
+    function proposeGlobalExitRoot(bytes32 proposedGlobalExitRoot) external {
```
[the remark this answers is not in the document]
invocamanman: i think it's a valid case :thinking: i could ban it, but, this way would be clear that there's a software bug!?

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,311 @@
...
+
+    // Number of reports that must match to consolidate a new rewards root (N/M)
+    uint64 public quorum;
+
+    // Oracle member address --> current voted GER
+    mapping(address => bytes32) public addressToLastProposedGER;
+
+    // GER --> Report(votes)
+    mapping(bytes32 => Report) public proposedGERToReport;
+
```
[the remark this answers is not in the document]
invocamanman: i'm unsure which verison shoudl i put here.
Shoudl we have a smart contract versioning?¿ should we follow the agglayer one?¿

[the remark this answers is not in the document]
krlosMata: Discussed internally. Version will be added with no interface to be forced until a common/proper interface is decided

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,311 @@
...
+    /**
```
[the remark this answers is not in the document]
krlosMata: agree

contracts/v2/sovereignChains/AggOracleCommittee.sol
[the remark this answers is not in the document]
krlosMata: Discussed internally to handle the edge case on the function. And given the scenario about lowering the case, a new function will be added that can consolidte a GER if quorum is reach

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,311 @@
...
+        address _newGlobalExitRootUpdater
+    ) external onlyOwner {
+        globalExitRootManagerL2Sovereign.transferGlobalExitRootUpdater(
+            _newGlobalExitRootUpdater
+        );
+    }
+
+    /**
+     * @notice Accept the globalExitRootUpdater role.
+     */
```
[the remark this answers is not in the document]
krlosMata: I think it is worth it to add the two of them. Not strong opinion here since it is only a comment xD

ignasirv (changes requested):
There is still missing the logic whan you want to run the AggOracleComittee from the docker script, when the `updateVanillaGenesis` genesis is called from `4_createRollup.ts` and `createNewRollup.ts`. This second one (the create new rollup tool) maye there is no need because if I am not wrong its only to get the genesis root which is not needed for new ALGateway/PP chains

tools/createSovereignGenesis/create-sovereign-genesis.ts
```diff
@@ -76,8 +76,64 @@ async function main() {
         });
     }
 
-    // check timelock parameters
+    if (createGenesisSovereignParams.useAggOracleCommittee === true) {
+        if (
```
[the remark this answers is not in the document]
invocamanman: i though the error would be more clear if it was on same file :thinking:

deployment/v2/utils/updateVanillaGenesis.ts
```diff
@@ -290,6 +288,70 @@ async function updateVanillaGenesis(genesis, chainID, initializeParams) {
...
+         * deploy aggORacle proxy and initialize
+         */
+        const transparentProxyFactory = await ethers.getContractFactory(
+            '@openzeppelin/contracts4/proxy/transparent/TransparentUpgradeableProxy.sol:TransparentUpgradeableProxy',
+        );
+
+        // create initialize transaction
+        // Initialize aggOracle Manager
+        const initializeAggOracleCommittee = aggOracleCommitteeFactory.interface.encodeFunctionData('initialize', [
+            proxiedTokensManager, // TODO
```
ignasirv, line 327: I think this is a good point to discuss, if using the `proxiedTokensManager` or a different address being set in the params.
If we set the `proxiedTokensManager` we will can not create sovereign chains with the token upgradeability "feature" disabled.

tools/createSovereignGenesis/create-sovereign-genesis.ts
```diff
@@ -76,8 +76,64 @@ async function main() {
...
+
+        if (createGenesisSovereignParams.quorum > createGenesisSovereignParams.aggOracleCommittee.length) {
+            logger.error(
+                `quorum must be smaller or equal than the number of aggOracleCommittee members (${createGenesisSovereignParams.aggOracleCommittee.length})`,
+            );
+            process.exit(1);
+        }
+    } else {
+        if (
+            createGenesisSovereignParams.globalExitRootUpdater === undefined ||
```
ignasirv, line 127: Checking `isAddress` already includes the undefined check

deployment/v2/utils/updateVanillaGenesis.ts
```diff
@@ -290,6 +288,70 @@ async function updateVanillaGenesis(genesis, chainID, initializeParams) {
...
+            initializeParams.aggOracleCommittee,
+            initializeParams.quorum,
+        ]);
+
+        const deployAggOracleProxy = (
+            await transparentProxyFactory.getDeployTransaction(
+                aggOracleImplementationAddress as string, // must have bytecode
+                proxiedTokensManager as string,
```
ignasirv, line 335: I won't set the `proxiedTokensManager` as the proxyOwner, maybe the ProxyAdmin?

deployment/v2/utils/updateVanillaGenesis.ts
```diff
@@ -290,6 +288,70 @@ async function updateVanillaGenesis(genesis, chainID, initializeParams) {
...
+                initializeAggOracleCommittee,
+            )
+        ).data;
+
+        // Update injectedTx to initialize GER
```
ignasirv, line 340:
```suggestion
        // Update injectedTx to initialize AggOracleProxy
```

deployment/v2/utils/updateVanillaGenesis.ts
invocamanman: i can add a new address to be the owner. AS i though we can always transfer it  aaand this should be a mutisig.
We can update it tho

invocamanman:
i will add the new address and the storage checks

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,344 @@
...
+        uint256 oracleMemberIndex
+    ) external onlyOwner {
+        bytes32 lastVotedReportHash = addressToLastProposedGER[
+            oracleMemberAddress
+        ];
+
+        require(lastVotedReportHash != bytes32(0), WasNotOracleMember());
+
+        require(
+            aggOracleMembers[oracleMemberIndex] == oracleMemberAddress,
```
krlosMata, line 245: if `oracleMemberIndex` is greater or equal than the array length, solidity will trigger a revert and will not hit return teh revert reason `OracleMemberIndexMismatch()`.
Addition check should be done in order to return a meaningful revert reason if `oracleMemberIndex  >= aggOracleMembers.length`

krlosMata (changes requested):

contracts/v2/sovereignChains/AggOracleCommittee.sol
invocamanman: oki will fix

invocamanman:
done ^^

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,349 @@
...
+ */
+contract AggOracleCommittee is IAggOracleCommittee, OwnableUpgradeable {
+    /**
+     * @notice Struct to store votes for GERs
+     * @param votes Current number of votes for this report
+     * @param timestamp Timestamp when the report was first proposed
+     */
+    struct Report {
+        uint64 votes;
+        uint64 timestamp;
```
web3security, line 21: timestamp is just used to check whether or not a Report does exist, just wondering if this is intended for future use like having a timeout or something or even controlling whether a new oracle member "just added" is allowed or not to vote for a specific ongoing voting (for example checking when he was added), because we can simply check if votes > 0 to check if the Report does exist.

It would be nice to have more context around this.

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,349 @@
...
+
+        // Check if it reaches the quorum
+        if (currentVotedReport.votes >= quorum) {
+            _consolidateGlobalExitRoot(proposedGlobalExitRoot);
+        } else {
+            // Store submitted report with a new added vote
+            proposedGERToReport[proposedGlobalExitRoot] = currentVotedReport;
+
+            // Store voted report hash
+            addressToLastProposedGER[msg.sender] = proposedGlobalExitRoot;
```
web3security, line 161: A voter reaching quorum with its first vote is not updating its addressToLastProposedGER, this seems to be expected however we should probably add a comment.

For example if the same voter is the one making the quorum for every vote, if will always keep its `lastProposedGER == INITIAL_PROPOSED_GER` instead of keeping track of the GER voted.

Changing it will change the flow and assumptions so I'd suggest at least to add a comment about this behavior in case it is expected.

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,349 @@
...
+        emit AddAggOracleMember(newOracleMember);
+    }
+
+    /**
+     * @notice Remove an oracle member.
+     * Only the owner can call this function.
+     * @param oracleMemberAddress Address of the oracle member to remove
+     * @param oracleMemberIndex Index of the oracle member to remove
+     */
+    function removeOracleMember(
```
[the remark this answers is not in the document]
invocamanman: We give full right to the owner

tools/createSovereignGenesis/create-sovereign-genesis.ts
```diff
@@ -21,6 +21,7 @@ import { GENESIS_CONTRACT_NAMES } from '../../src/utils-common-aggchain';
 // read files
 import genesisBase from './genesis-base.json';
 import createGenesisSovereignParams from './create-genesis-sovereign-params.json';
+import { create } from 'domain';
```
ignasirv, line 24: Remove

contracts/v2/sovereignChains/AggOracleCommittee.sol
krlosMata: It could also add metadata information about when the votes has been signed.
Given that this is on L2 and the GAS cost is not relevant

invocamanman: since the gas cost of this is extremly minimal ( since it does nto read/store an additional time the storage )
Adds temporary metadata and makes easier future upgrades we agree that for that minimal cost is worth let it.

deployment/v2/utils/updateVanillaGenesis.ts
```diff
@@ -290,6 +292,73 @@ async function updateVanillaGenesis(genesis, chainID, initializeParams) {
...
+                txObjectDeployAggOracleImpl.unsignedHash,
+                txObjectDeployAggOracleImpl.signature as any,
+            ),
+        );
+
+        const txDeployAggORacleCommitee = processorUtils.rawTxToCustomRawTx(txObjectDeployAggOracleImpl.serialized);
+        batch2.addRawTx(txDeployAggORacleCommitee);
+        aggOracleImplementationAddress = getContractAddress({
+            from: txObjectDeployAggOracleImpl.from,
+            nonce: injectedTx.nonce,
```
web3security, line 317: can we review this nonce? Shouldn't we use `txObjectDeployAggOracleImpl.nonce` instead of `injectedTx.nonce` ?

deployment/v2/utils/updateVanillaGenesis.ts
```diff
@@ -290,6 +292,73 @@ async function updateVanillaGenesis(genesis, chainID, initializeParams) {
...
+                proxyAdminAddress as string,
+                initializeAggOracleCommittee,
+            )
+        ).data;
+
+        // Update injectedTx to initialize AggOracleProxy
+        injectedTx.to = null;
+        injectedTx.data = deployAggOracleProxy;
+
+        const txObject2 = ethers.Transaction.from(injectedTx);
```
web3security, line 351: overlapping with other different `txObject2` defined below, recommended to use different naming and probably use a naming less generic.

tools/deployAggOracleCommittee/deployAggOracleCommittee.ts
```diff
@@ -0,0 +1,123 @@
...
+    expect(membersCount).to.be.equal(BigInt(aggOracleMembers.length));
+
+    // Compute output
+    const outputJson = {
+        gitInfo: getGitInfo(),
+        aggOracleCommitteeAddress: aggOracleCommitteeContract.target,
+        deployer: deployer.address,
+        proxyAdminAddress: proxyAdmin.target,
+        proxyOwnerAddress,
+        ownerAddress,
```
web3security, line 108: Recommended to use better naming for all "owner" related roles, ownerAddress is proxyAdmin.owner, AggOracleCommittee Proxy.owner is proxyAdmin and "owner" is the one for onlyOwner modifier.. right? It is a bit confusing.

tools/deployAggOracleCommittee/deploy_parameters.json.example
```diff
@@ -0,0 +1,14 @@
+{
+    "globalExitRootManagerL2SovereignAddress": "0x0000000000000000000000000000000000000000",
+    "ownerAddress": "0x0000000000000000000000000000000000000000",
```
web3security, line 3: ownable is protecting us from an address(0) invalid assignment but better to avoid address(0)

tools/deployAggOracleCommittee/deploy_parameters.json.example
```diff
@@ -0,0 +1,14 @@
+{
+    "globalExitRootManagerL2SovereignAddress": "0x0000000000000000000000000000000000000000",
```
web3security, line 2: The constructor is not protected against an address(0) assignment for globalExitRootManagerL2SovereignAddress, I'd suggest we add it.

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,353 @@
...
+    mapping(address => bytes32) public addressToLastProposedGER;
+
+    // GER --> Report(votes)
+    mapping(bytes32 => Report) public proposedGERToReport;
+
+    /**
+     * @notice Disables initializers on the implementation, following best practices.
+     */
+    constructor(IGlobalExitRootManagerL2SovereignChain globalExitRootManager) {
+        globalExitRootManagerL2Sovereign = globalExitRootManager;
```
web3security, line 52: Recommended to add protection against an invalid address(0) assignment as a best practice. 

This is especially interesting as template json data does have address(0).

deployment/v2/utils/updateVanillaGenesis.ts
invocamanman: they are the same, but y will update

contracts/v2/sovereignChains/AggOracleCommittee.sol
```diff
@@ -0,0 +1,349 @@
...
+    /**
+     * @notice Update the quorum value.
+     * Only the owner can call this function.
+     * It is expected that the quorum and the oracle member will be updated atomically from a smart contract.
+     * @param newQuorum New quorum value
+     */
+    function updateQuorum(uint64 newQuorum) external onlyOwner {
+        require(newQuorum != 0, QuorumCannotBeZero());
+
+        quorum = newQuorum;
```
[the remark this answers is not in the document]
web3security: My concern is the edge case described above where GER-A is meant to be finally **wrong** and GER-B is the one to be consolidated (while in the process of moving all votes from GER-A to GER-B).

The owner will have to be careful while updating quorum if there are ongoing votes like in this edge case that should never be consolidated especially because consolidateGlobalExitRoot has no access control in place.

I agree that being a privileged function it is an edge case, my suggestion is that the owner should control any ongoing voting while doing this operations to resume & consolidate specific GERs avoiding the consolidation of invalid ones.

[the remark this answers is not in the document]
invocamanman: yee good point, i think in that case, the best way to proceed will be that the oracles vote the new GERB before updating the quorum :thinking:

krlosMata (approved):
😸

PR #501, opened 2025-07-14
fix: small audit informational remediation

ignasirv:
Update to equivalence condition but more readable following recommendations from audit informational finding https://cantina.xyz/code/66c3b4bc-9068-4001-a3f4-2dec2385965a/findings/13

### DIFF
--- a/contracts/v2/PolygonZkEVMBridgeV2.sol
+++ b/contracts/v2/PolygonZkEVMBridgeV2.sol
@@ -65,7 +65,7 @@ contract PolygonZkEVMBridgeV2 is
     uint256 internal constant _GLOBAL_INDEX_MAINNET_FLAG = 2 ** 64;
 
     // Current bridge version
-    string public constant BRIDGE_VERSION = "v1.0.0";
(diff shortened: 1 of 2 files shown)


contracts/v2/PolygonZkEVMBridgeV2.sol
```diff
@@ -440,7 +440,7 @@ contract PolygonZkEVMBridgeV2 is
         bytes calldata metadata
     ) external payable ifNotEmergencyState {
         // If exist a gas token, only allow call this function without value
-        if (msg.value != 0 && address(WETHToken) != address(0)) {
+        if (msg.value != 0 && gasTokenAddress != address(0)) {
```
krlosMata, line 443: Latest conversations around outpost may add the possibility to identify native gasToken as:
- `gasTokenNetwork: address(0)`
- `gasTokenNetwork: X`

Therefore, the check is incomplete. Althought it is not yet supported in the PP I would identify ETH as:
- `gasTokenNetwork != address(0) && gasTokenNetwork != 0`

krlosMata (changes requested):

PR #502, opened 2025-07-18
add versions AggLayerGateway & AggchainECDSA

laisolizq:


contracts/v2/AggLayerGateway.sol
```diff
@@ -37,6 +37,9 @@ contract AggLayerGateway is
     bytes32 internal constant AL_FREEZE_PP_ROUTE_ROLE =
         keccak256("AL_FREEZE_PP_ROUTE_ROLE");
 
+    // Current AggLayerGateway version
+    string public constant AGGLAYERGATEWAY_VERSION = "al-v0.3.0";
```
ignasirv, line 41:
```suggestion
    string public constant AGGLAYER_GATEWAY_VERSION = "al-v0.3.0";
```

contracts/v2/aggchains/AggchainECDSA.sol
```diff
@@ -22,6 +22,10 @@ contract AggchainECDSA is AggchainBase {
     // Aggchain type selector, hardcoded value used to force the last 2 bytes of aggchain selector to retrieve  the aggchain verification key
     bytes2 public constant AGGCHAIN_TYPE = 0;
 
+    /// @notice Op L2OO Semantic version.
+    /// @custom:semver v2.0.0
+    string public constant version = "v2.0.0";
```
krlosMata, line 27: Let's add our own semantic version like: `AGGCHAIN_ECDSA_VERSION: "v0.0.1"`

For example, this SC is not on production. Therefore, the major release is 0.

contracts/v2/AggLayerGateway.sol
krlosMata: I'd suggest to follow our own SC versioning, not dependant on the `al`.
Therefore, I'd set `v1.0.0` since the SC is on production and has not been yet modified

laisolizq:
## Steps

- You make the change(s) in the necessary contract(s).
- You evaluate the change:
    - Is it a fix or a comment? → `X.Y.Z+1`
    - Is it a new feature? → `X.Y+1.Z`
    - Is it a breaking change? → `X+1.Y.Z`
- You update the contract version.
- You create a tag with the same change as before, but for the repository version, and add `rc.1`.
- You will continue iterating with changes, incrementing to `rc.X+1`.
- When it’s finished and you are ready to do a `release` for that iteration:
    - You can generate the changelog with:
    `npx tsx docs/generate_changelog.ts <tag of the last release in the changelog file>`
    - Review the generated data (if everything works correctly, the release version in the changelog should match what you previously calculated).
    - Complete the explanations for the commits.
    - Upload the `CHANGELOG.md`.
    - Create the `tag` and the `release`.

laisolizq:
More info: https://github.com/agglayer/protocol-team-kanban/issues/558

contracts/v2/PolygonRollupManager.sol
```diff
@@ -237,7 +237,7 @@ contract PolygonRollupManager is
         keccak256("EMERGENCY_COUNCIL_ADMIN");
 
     // Current rollup manager version
-    string public constant ROLLUP_MANAGER_VERSION = "al-v0.3.1";
+    string public constant ROLLUP_MANAGER_VERSION = "v2.1.0";
```
ignasirv, line 240: Why not starting with version v1.0.0?

laisolizq: As I say in the comment on this PR:
> For contracts that previously had the version al-v0.3.0 and already existed, I’ve chosen v2.0.0 (also because of the v2 folder they are located in).

I think it makes more sense, but open to suggestions!^

ignasirv: I would use v1.0.0, I guess in a future clean of the contracts folders the v2 wil be removed.

krlosMata: I think `v1` fits better here.
At the end, we are now doing just a mapping from `al-v0.3.1` to `vX.Y.Z`.
Since it is the first version with the new versioning, `v1` may fit better here.
The folder name should be agnostic to the versioning.

contracts/v2/aggchains/AggchainFEP.sol
```diff
@@ -56,7 +56,7 @@ contract AggchainFEP is AggchainBase {
 
     /// @notice Op L2OO Semantic version.
     /// @custom:semver v2.0.0
-    string public constant version = "v2.0.0";
+    string public constant AGGCHAIN_FEP_VERSION = "v1.0.0";
```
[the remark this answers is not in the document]
laisolizq: For `version`? Or `v2.0.0`?

docs/CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
+
+> This CHANGELOG is a bit different; we are only adding the versions for the first time.
+> In the Breaking Changes section, we’ve listed the versions that have been changed.
+> In the New Features section, we’ve listed the versions that didn’t exist before.
+
+## [v11.0.0] - 2025-07-21 (First ChangeLog)
+
+### 🚨 Breaking Changes
```
alrevuelta, line 8: are these really breaking changes?

laisolizq: Probably not! But I think it's important to emphasize that we've changed the version so that everyone knows.

krlosMata (approved):
😸

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -18,7 +18,7 @@ contract BridgeL2SovereignChain is
     using SafeERC20 for ITokenWrappedBridgeUpgradeable;
 
     // Current bridge version
-    string public constant BRIDGE_SOVEREIGN_VERSION = "v10.1.2";
+    string public constant BRIDGE_SOVEREIGN_VERSION = "v1.0.0";
```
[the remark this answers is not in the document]
krlosMata: discussed AFK the following:
- first phase to leave it with current versioning name
- second phase to add a common interface `version()` for each contract

Second phase to be done in the upcoming release given all the current changes in parallel that are being done

PR #509, opened 2025-07-30
Add IVersion interface

laisolizq:
- Add IVersion interface
- Unify `version` function

### DIFF
--- a/contracts/v2/aggchains/AggchainECDSA.sol
+++ b/contracts/v2/aggchains/AggchainECDSA.sol
@@ -241,7 +241,7 @@ contract AggchainECDSA is AggchainBase {
      * @notice Function to retrieve the current version of the contract.
      * @return version of the contract.
      */
-    function version() external view override returns (string memory) {
+    function version() external pure returns (string memory) {
         return AGGCHAIN_ECDSA_VERSION;
     }
 }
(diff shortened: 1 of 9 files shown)


contracts/v2/aggchains/AggchainECDSA.sol
```diff
@@ -236,4 +236,12 @@ contract AggchainECDSA is AggchainBase {
         // Emit event
         emit OnVerifyPessimisticECDSA(newStateRoot);
     }
+
+    /**
+     * @notice Function to retrieve the current version of the contract.
+     * @return version of the contract.
+     */
+    function version() external view override returns (string memory) {
```
ignasirv, line 244: All this version function that are reading the version from a constant (bytecode) can be moved to `pure` and no need to add the `override` if they are not inheriting from a virtual funcion

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -1066,4 +1066,12 @@ contract BridgeL2SovereignChain is
             _increaseLocalBalanceTree(_MAINNET_NETWORK_ID, address(0), amount);
         }
     }
+
+    /**
+     * @notice Function to retrieve the current version of the contract.
+     * @return version of the contract.
+     */
+    function version() external pure override returns (string memory) {
```
invocamanman, line 1074: same comment as ignasi said, override not necessary

PR #513, opened 2025-08-14
add configs

invocamanman:
- Add everything regarding having multiple configurations to choose via agglayer
- If not version is selected, last parameters are used, 100% compatible

TODO:
- succint configs can be erased, therefore, the opSuccint approach should be updated saving the previous configs


invocamanman:
this feature will remain on hold

PR #519, opened 2025-09-03
Fix found informationals

ignasirv:
Fixed following informationals from https://github.com/agglayer/internal-audit/issues/184

- fix comment in IBridgeL2SovereignChains:InvalidLBTLeaf error
- keep permit comments present in PolygonZkEVMBridgeV2V1010.sol:_permit:1185 & PolygonZkEVMBridgeV2V1010.sol:_permit:1243
- in BridgeLib:returnDataToString is an internal function. It should be renamed as the _returnDataToString
- in DepositContractBase:_isValidSubtreeFrontier ensure that not matched siblings are zero (just to not set random values to the new frontier)
- in BridgeL2SovereignChain.sol --> events BackwardLET & ForwardLET add also the previousDepositCount& previusRoot
- in BridgeL2SovereignChain.sol:763 --> it is not the GER, is the LER. Rename to avoid confiusion.
- in BridgeL2SovereignChain.sol:779 --> it is not a stateRoot, it is a LER
- modify also error name InvalidExpectedRoot to be more explicit (like 'InvalidExpectedLER')
- fix all references to stateRoot
- in BridgeL2SovereignChain.sol:setMultipleClaims() the globalIndex is checked almost as it is checked already on the PolygonZkEVMBridgeV2.sol:_verifyLeaf -> why not create an internal function and reuse the the code to rebuild the globalIndex ?
- in BridgeL2SovereignChain.sol:setLocalBalanceTree() the originNetwork[i] == networkIDI think it is not necessary. The LBT in the SC can be corrupted at some specific leave, regardless the originNetwork

### DIFF
--- a/contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
+++ b/contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
@@ -182,17 +182,31 @@ contract BridgeL2SovereignChain is
 
     /**
      * @dev Emitted when local exit tree is moved backward
+     * @param previousDepositCount The deposit count before moving backward
+     * @param previousRoot The root of the local exit tree before moving backward
      * @param newDepositCount The resulting deposit count after moving backward
      * @param newRoot The resulting root of the local exit tree after moving backward
      */
-    event BackwardLET(uint256 newDepositCount, bytes32 newRoot);
+    event BackwardLET(
+        uint256 previousDepositCount,
+        bytes32 previousRoot,
+        uint256 newDepositCount,
+        bytes32 newRoot
+    );
 
     /**
      * @dev Emitted when local exit tree is moved forward
+     * @param previousDepositCount The deposit count before moving forward
+     * @param previousRoot The root of the local exit tree before moving forward
      * @param newDepositCount The resulting deposit count after moving forward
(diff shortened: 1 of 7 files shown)


contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
laisolizq, line 1: I think we could update the constructor’s comment by documenting the assignment of the deployer

contracts/v2/sovereignChains/BridgeL2SovereignChain.sol
```diff
@@ -609,35 +623,12 @@ contract BridgeL2SovereignChain is
...
-                // Reconstruct global index to assert that all unused bits are 0
-                require(
-                    _GLOBAL_INDEX_MAINNET_FLAG + uint256(leafIndex) ==
-                        globalIndex,
-                    InvalidGlobalIndex()
-                );
-            }
+            // Validate and decode global index using shared logic
+            (
+                uint32 leafIndex, // indexRollup
```
krlosMata, line 628: comment `// indexRollup` should be in the below line

krlosMata (changes requested):

krlosMata (changes requested):

PR #561, opened 2025-11-05
Remove boilerplate code in tests using common consts

alrevuelta:
* This PR moves repeated consts related to roles in the tests into a common file.

Notes:
* There are more usages of these consts beyond the tests. But want to keep the PR low risk. So no touching anything beyond the tests.
* Instead of importing `from '../helpers/consts';` one by one, we could import all `*` as `consts` and then use `consts.XXX` but wanted to keep the diff minimal.


jesusholoo (approved):

krlosMata (commented):
Maybe we can place it in `src/constants.ts`....in case those constants wants to be used from tooling.
I think it makes more sense

krlosMata:
linter

alrevuelta:
@krlosMata 

> Maybe we can place it in src/constants.ts....in case those constants wants to be used from tooling.
I think it makes more sense

indeed. didnt know about that file. fixed: 3315b8932f62e7b0c9b5b8c27dce84474ffa7a04

krlosMata (approved):
😸

PR #567, opened 2025-11-11
Feature/bridge ger v4

laisolizq:
- Update precoms AgglayerBridgeL2 and AgglyarGERL2
- Opt bytecode AgglyaerBridgeL2


contracts/sovereignChains/AgglayerGERL2.sol
```diff
@@ -217,6 +253,149 @@ contract AgglayerGERL2 is
...
+        if (newLocalExitRoots.length != networkIDs.length) {
+            revert InputArraysLengthMismatch();
+        }
+        // do not insert LER if already set
+        for (uint256 i = 0; i < newLocalExitRoots.length; i++) {
+            bytes32 keyLER = getHashLER(newLocalExitRoots[i], networkIDs[i]);
+            if (localExitRootMap[keyLER] == false) {
+                localExitRootMap[keyLER] = true;
+                // Update hash chain value
+                insertedLERHashChain = Hashes.efficientKeccak256(
```
alrevuelta: maybe here you can follow a similar approach to `removeLERs`? I mean, accumulate to a local variable `nextInsertedLERHashChain` and then assign it to `insertedLERHashChain` out of the loop. That should be a single SSTORE instead multiple.

contracts/sovereignChains/AgglayerGERL2.sol
```diff
@@ -217,6 +253,149 @@ contract AgglayerGERL2 is
...
+            smtProofLocalExitRoots.length != originNetworks.length ||
+            smtProofLocalExitRoots.length != originTokenAddresses.length ||
+            smtProofLocalExitRoots.length != destinationNetworks.length ||
+            smtProofLocalExitRoots.length != destinationAddresses.length ||
+            smtProofLocalExitRoots.length != amounts.length ||
+            smtProofLocalExitRoots.length != metadatas.length
+        ) {
+            revert InputArraysLengthMismatch();
+        }
+        if (localExitRoots.length != 1 || networkIDs.length != 1) {
```
alrevuelta: not sure i understand this. if length has to be 1, why is the type an array []?

PR #569, opened 2025-11-12
script push docker all release

laisolizq:
Add new script to push `dockerv2:contracts:all`

### DIFF
--- a/.github/workflows/build-push-docker-all.yml
+++ b/.github/workflows/build-push-docker-all.yml
@@ -1,0 +1,138 @@
+name: Push Docker image
+
+on:
+  release:
+    types: [created]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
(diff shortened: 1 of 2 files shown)


.github/workflows/build-push-docker-all.yml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: Prepare
+        run: |
+          platform=${{ matrix.platform }}
+          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: setup
```
alrevuelta, line 43: out of curiosity, is it needed to install docker compose?

krlosMata (approved):
😸

.github/workflows/build-push-docker-all.yml
laisolizq: yes! We use it with this file: https://github.com/agglayer/agglayer-contracts/blob/feature/v12/docker/docker-compose.yml

PR #571, opened 2025-11-13
ObsoleteRollup tool

krlosMata:
This PR does the following:
- add new tool `obsoleteRollupType`
  - allows inclusion or exclusion list

### DIFF
--- a/tools/obsoleteRollupType/README.md
+++ b/tools/obsoleteRollupType/README.md
@@ -1,0 +1,103 @@
+# Obsolete rollup type
+Script to call `obsoleteRollupType` function in the `AgglayerManager.sol` smart contract.
+
+## Install
+```
+npm i
+```
+
+## Setup
+### obsoleteRollupType
+The script supports two modes of operation:
+
+#### Mode 1: Direct List (obsoleteRollupTypes)
+Specify exactly which rollup types to obsolete.
+
+#### Mode 2: Exclusion List (excludedRollupTypesID)
+Specify which rollup types to keep active; all others will be obsoleted.
+
+**Note**: You must use either `obsoleteRollupTypes` OR `excludedRollupTypesID`, not both.
+
+#### Config file parameters:
+  - `type`: Specify the type of transaction execution:
+    - `EOA`: If executing from a wallet, the script will execute the obsolete rollup type operation on the specified network
+    - `Multisig`: If executing from a multisig, the script will output the calldata of the transaction to execute
+    - `Timelock`: If executing through a timelock, the script will output the execute and schedule data to send to the timelock contract
+  - `agglayerManagerAddress`: `AgglayerManager.sol` SC address
+  - `timelockDelay (optional)`: at least it should be the minimum delay of the timelock smart contract
+  - `timelockSalt (optional)`: timelock salt
+  - `maxFeePerGas`: set custom gas
+  - `maxPriorityFeePerGas`: set custom gas
+  - `multiplierGas`: set custom gas
+  - **EITHER** `obsoleteRollupTypes` (Mode 1): array of rollup type IDs to obsolete
+    - Each element is a `rollupTypeID` (number) that will be marked as obsolete
+  - **OR** `excludedRollupTypesID` (Mode 2): array of rollup type IDs to **exclude** from being obsoleted
+    - The script will automatically fetch all rollup types from the AgglayerManager
+    - It will exclude rollup types that are already obsolete
+    - It will exclude rollup types in the `excludedRollupTypesID` array
+    - All remaining rollup types will be marked as obsolete
+- A network should be selected when running the script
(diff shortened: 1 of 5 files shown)


krlosMata:
@codex review

jesusholoo:
think a cool feature will be "purge" unused rollups.
 Just a bool, loops thorugh all rollups, all types cuurrently unused, if not obsolete, will be obsolete

krlosMata:
> think a cool feature will be "purge" unused rollups. Just a bool, loops through all rollups, all types cuurrently unused, if not obsolete, will be obsolete

cool !! I think it is worth it xD
There will be three modes: `inclusion`, `exclusion` and `purge`

Also, I'd add support for `MultiSendCallOnly`.

tools/obsoleteRollupType/README.md
```diff
@@ -0,0 +1,104 @@
...
+Specify exactly which rollup types to obsolete.
+
+#### Mode 2: Exclusion List (excludedRollupTypesID)
+Specify which rollup types to keep active; all others will be obsoleted.
+
+**Note**: You must use either `obsoleteRollupTypes` OR `excludedRollupTypesID`, not both.
+
+#### Config file parameters:
+  - `type`: Specify the type of transaction execution:
+    - `EOA`: If executing from a wallet, the script will execute the obsolete rollup type operation on the specified network
```
alrevuelta, line 33: in the eoa case how do you load the private key that is used to call the tx? some env? cant find it in the docs.

tools/obsoleteRollupType/obsoleteRollupType.ts
```diff
@@ -0,0 +1,275 @@
...
+        const outputJson = {} as any;
+        const rollupTypeID = rollupTypesToObsolete[i];
+
+        outputJson.networkName = network.name;
+        outputJson.agglayerManagerAddress = agglayerManagerAddress;
+        outputJson.rollupTypeID = rollupTypeID;
+
+        if (obsoleteRollupTypesParameters.type === transactionTypes.EOA) {
+            // Check role
+            if ((await rollupManagerContract.hasRole(OBSOLETE_ROLLUP_TYPE_ROLE, deployer.address)) === false) {
```
alrevuelta: this if (hasRole) can be run just once out of the loop?

tools/obsoleteRollupType/README.md
krlosMata: adding docs

tools/obsoleteRollupType/obsoleteRollupType.ts
krlosMata: yep!

tools/obsoleteRollupType/README.md
```diff
@@ -0,0 +1,213 @@
...
+## Setup
+### obsoleteRollupType
+The script supports three modes of operation, controlled by the `mode` parameter:
+
+#### Mode 1: Inclusion (`mode: "inclusion"`)
+Specify exactly which rollup types to obsolete using the `list` parameter.
+- Use case: You know the specific rollup type IDs that need to be obsoleted
+- Requires: `list` parameter with array of rollup type IDs
+
+#### Mode 2: Exclusion (`mode: "exclusion"`)
```
alrevuelta, line 18: maybe a bit opinionated but imho the modes 2) and 3) should not exist.
* they add cognitive load to a tool that is rather simple. just look at the docs and code. tons of cases. we can cut that by /3.
* better to be explicit. mode 1 is explicit. you state exactly what you want to obsolete.
* mode 2 is not explicit. easy to mess up and end up doing the opposite of what you wanted. i dont think the time it saves justifies the complexity. if you have 1, 2, 3, 4, 5 and want to obsolete 2, 3, 4, 5 just use mode 1 with 2, 3, 4, 5.

just trying to prevent obsoleting something by accident. dangerous because it doesnt seem to be a way to revert that.

alrevuelta (commented):
left a comment re mode 2 and 3. not sure the complexity they add in the code justify the value they add. better to be explicit and obsolete exactly what you want.

tools/obsoleteRollupType/README.md
krlosMata: It is a fair point indeed.
I think it adds some value since otherwise we would use Mode_1 and generate manually the lists of rollup to obsolete. In mode_2, mode_3...the list is created automatically by the script, rather than externally, avoiding errors on external scripts or human errors

krlosMata: @alrevuelta is it ok as it is ?
I was thinking on just not allowing `purge` with an EOA address mode

alrevuelta: not blocking. just a preference for safety.

alrevuelta (approved):

PR #576, opened 2025-11-26
Cleanup docker compose

alrevuelta:
* There were two almost identical scripts [deploy-docker.sh](https://github.com/agglayer/agglayer-contracts/blob/feature/v12/docker/scripts/v2/deploy-docker.sh) and [deploy-dockerv2.sh](https://github.com/agglayer/agglayer-contracts/blob/feature/v12/docker/scripts/v2/deploy-dockerv2.sh).
* The only difference is that one uses "docker compose" and the other "docker-compose". This is due to a change in docker compose version.
* This PR removes one script, since they are not needed and makes a small modification so that a single script works with both docker compose version. With "-" and without "-".


krlosMata:
This cleanup will come with v13, I'd suggest to change the PR to `draft`

PR #610, opened 2026-07-28
feat: add ERC-7786 gateway adapter for AgglayerBridge

taylanpince:
Add a standalone ERC-7786 gateway adapter that wraps the AgglayerBridge message-passing layer, with ERC-7930 interoperable address encoding, a deploy tool for L1/L2, and standalone tests.

### DIFF
--- a/contracts/periphery/erc7786/AgglayerERC7786Gateway.sol
+++ b/contracts/periphery/erc7786/AgglayerERC7786Gateway.sol
@@ -1,0 +1,309 @@
+// SPDX-License-Identifier: AGPL-3.0
+
+pragma solidity 0.8.28;
+
+import {Ownable} from "@openzeppelin/contracts5/access/Ownable.sol";
+import {IBridgeMessageReceiver} from "../../interfaces/IBridgeMessageReceiver.sol";
+import {IERC7786GatewaySource, IERC7786Recipient} from "./IERC7786.sol";
+import {InteroperableAddress} from "./InteroperableAddress.sol";
+
+/**
+ * @dev Minimal interface of the AgglayerBridge required by the gateway
+ */
+interface IAgglayerBridgeMessenger {
+    function bridgeMessage(
+        uint32 destinationNetwork,
+        address destinationAddress,
+        bool forceUpdateGlobalExitRoot,
+        bytes calldata metadata
+    ) external payable;
+
+    function networkID() external view returns (uint32);
+}
+
+/**
+ * @title AgglayerERC7786Gateway
+ * @notice ERC-7786 gateway adapter on top of the AgglayerBridge message-passing layer.
+ * See https://eips.ethereum.org/EIPS/eip-7786
+ *
+ * Acts as a source gateway: `sendMessage` wraps the message into the AgglayerBridge
+ * `bridgeMessage` flow, targeting the trusted counterpart gateway on the destination
+ * network. Acts as a destination gateway: `onMessageReceived` (invoked by the bridge
+ * on claim) validates the origin and forwards the message to the final
+ * `IERC7786Recipient`.
+ *
+ * ERC-7930 interoperable addresses use eip155 (EVM) chain ids. The owner registers
+ * each remote network with a single call to `registerRemoteNetwork`, which maps the
+ * chain id to the Agglayer network id and sets the trusted counterpart gateway
+ * deployed on that network.
+ */
+contract AgglayerERC7786Gateway is
+    IERC7786GatewaySource,
+    IBridgeMessageReceiver,
+    Ownable
+{
+    using InteroperableAddress for bytes;
+
+    /**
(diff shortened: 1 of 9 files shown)


contracts/periphery/erc7786/AgglayerERC7786Gateway.sol
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * @notice Entry point called by the AgglayerBridge when a message targeting this
+     * gateway is claimed. Validates the origin and delivers the message to the final
+     * ERC-7786 recipient, forwarding any native value.
+     * @param originAddress Address that called `bridgeMessage` on the origin network
+     * (must be the trusted counterpart gateway)
+     * @param originNetwork Agglayer network id of the origin network
+     * @param data Cross-chain metadata: abi.encode(sendId, sender, recipient, payload)
+     */
+    function onMessageReceived(
```
web3security, line 230: ### Cross-topology ETH value stranded as WETH in the gateway

When a value-bearing message is bridged from an **ETH-native source** to a **WETHToken (sovereign / non-ETH gas-token) destination**, `AgglayerBridge.claimMessage` takes the WETHToken path:

1. `_claimWrappedAsset(WETHToken, destinationAddress, amount)` → `WETHToken.mint(gateway, amount)`
2. `destinationAddress.call(onMessageReceived(...))` is invoked **without `{value}`**

So the value is minted as WETH **into the gateway**, and `onMessageReceived` forwards `msg.value = 0` to the recipient. `AgglayerERC7786Gateway` has no `receive()`, no ERC-20 forwarding and no rescue function → the WETH is **permanently stranded**.

The source-side guard (`AgglayerBridge.sol` ~L424) only rejects `msg.value != 0` when the **source** bridge holds a WETHToken. An ETH-native source → WETHToken destination bypasses it entirely.

**Impact:** any user who sends native ETH with a cross-chain message to a sovereign chain running a non-ETH gas token silently loses 100% of that value. The claim succeeds and emits `ClaimEvent`, so the loss is invisible to the sender and to event-based monitoring.

**PoC asserts:** `wethToken.balanceOf(gateway) == 3e18` (stranded), recipient WETH `== 0`, recipient ETH `== 0`, `recipient.lastValue() == 0` — while `ClaimEvent` is emitted.

**Suggested fix:** handle the WETHToken destination path for value-bearing messages — e.g. unwrap/forward the WETH (or native ETH) to the final recipient inside `onMessageReceived`, and/or add an ERC-20 rescue on the gateway. Add a regression test for the ETH-native → WETHToken topology.

Affected code:
- `contracts/bridge/AgglayerBridge.sol` — `claimMessage` WETHToken path (~L751-762) and source-side guard (~L424)
- `contracts/periphery/erc7786/AgglayerERC7786Gateway.sol` — `onMessageReceived` (no ERC-20 handling / no rescue)

contracts/periphery/erc7786/AgglayerERC7786Gateway.sol
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * @notice Registers a remote network: maps its EIP-155 chain id to its Agglayer
+     * network id (bidirectionally) and sets the trusted counterpart gateway deployed
+     * on that network
+     * @param chainId EIP-155 chain id of the remote network (must be non-zero)
+     * @param networkID Agglayer network id of the remote network
+     * @param gateway Counterpart gateway address on that network (must be non-zero)
+     */
+    function registerRemoteNetwork(
```
web3security, line 134: ### Stale `chainId → networkID` entry after `registerRemoteNetwork` re-registration

`registerRemoteNetwork(chainId, networkID, remoteGateway)` sets `_chainIdToNetworkID[newChainId]` and `_networkIDToChainId[networkID]` but **never clears the previous `_chainIdToNetworkID[oldChainId]`**. The stale entry keeps resolving `oldChainId → networkID` indefinitely.

Since dispatch resolves `chainId → networkID → remoteGateways[networkID]`, a sender that encodes the **old** chainId is **silently routed to the re-mapped gateway** instead of reverting.

**PoC:** register `(1101, netID 1, gatewayA)`, then re-register `(9999, netID 1, gatewayB)`. `chainIdToNetworkID(1101)` is still `1`, and `sendMessage` to chain `1101` emits `BridgeEvent` with `destinationAddress = gatewayB` (chain 9999) — `gatewayA` never receives it, no revert.

**Suggested fix:** in `registerRemoteNetwork`, before writing the new mapping, delete the stale reverse entry (`delete _chainIdToNetworkID[oldChainId]` when it points to this `networkID`) so a stale chainId resolves to unregistered and `sendMessage` reverts instead of misrouting.

Affected code:
- `contracts/periphery/erc7786/AgglayerERC7786Gateway.sol` — `registerRemoteNetwork` (~L142)

